Issue #41, opened 2021-09-18
Block rewards token accounting system tests


stewartie4:
Investigate and Implement system acceptance tests that test token accounting for block rewards

Example test cases are below but this is not a complete list.
Also create your own, thinking about security, param validation, error messaging, performance  etc.

When a block is mined:
- how are these rewards split between service providers?
- What wallet do these rewards come from and where do they go?

https://drive.google.com/file/d/1tNjb_OvX93s47QQ2gFrfeUBdmP0PcL9k/view will be an invaluable resource, also chat to the blockchain devs

Following the test style guide and the style of the existing tests

PR #75, opened 2021-10-16
Block rewards tests 

avanaur:
Resolves https://github.com/0chain/system_test/issues/41

Scenarios
- Miner rewards 
- Sharder rewards

No tests involve staking yet.

Logic for the tests are basically same.
- Get starting and ending balances of node (service provider) on time X and X+Y 
- Do some transactions in between to generate block fees
- For all rounds covered, calculate the block fees and block rewards and the expected share of the node (service provider)
- Assert the balance differences matches the calculated amount

The tests requires doing API calls for getting block/round details and getting node balances with round number associated.

### DIFF
--- a/tests/cli_tests/block_rewards_test.go
+++ b/tests/cli_tests/block_rewards_test.go
@@ -1,0 +1,378 @@
+package cli_tests
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"testing"
+
+	apimodel "github.com/0chain/system_test/internal/api/model"
+	climodel "github.com/0chain/system_test/internal/cli/model"
+	cliutil "github.com/0chain/system_test/internal/cli/util"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	blockRewardConfigKey       = "block_reward"
+	epochConfigKey             = "epoch"
+	rewardDeclineRateConfigKey = "reward_decline_rate"
+	rewardRateConfigKey        = "reward_rate"
+	shareRatioConfigKey        = "share_ratio"
+)
+
+func TestBlockRewards(t *testing.T) { // nolint:gocyclo // team preference is to have codes all within test.
+	t.Parallel()
+
+	t.Run("Miner rewards", func(t *testing.T) {
+		t.Parallel()
+
+		output, err := registerWallet(t, configPath)
+		require.Nil(t, err, "registering wallet failed", strings.Join(output, "\n"))
+
+		output, err = executeFaucetWithTokens(t, configPath, 1)
+		require.Nil(t, err, "faucet execution failed", strings.Join(output, "\n"))
+
+		// Get MinerSC Global Config
+		output, err = getMinerSCConfig(t, configPath)
+		require.Nil(t, err, "get miners sc config failed", strings.Join(output, "\n"))
+		require.Greater(t, len(output), 0)
+
+		mconfig := map[string]float64{}
+		for _, o := range output {
+			configPair := strings.Split(o, "\t")
+			val, err := strconv.ParseFloat(strings.TrimSpace(configPair[1]), 64)
+			require.Nil(t, err, "config val %s for %s is unexpected not float64: %s", configPair[1], configPair[0], strings.Join(output, "\n"))
+			mconfig[strings.TrimSpace(configPair[0])] = val
+		}
+
+		// Get miner list.
+		output, err = getMiners(t, configPath)
+		require.Nil(t, err, "get miners failed", strings.Join(output, "\n"))
+		require.Len(t, output, 1)
+
+		var miners climodel.NodeList
+		err = json.Unmarshal([]byte(output[0]), &miners)
+		require.Nil(t, err, "Error deserializing JSON string `%s`: %v", output[0], err)
+		require.NotEmpty(t, miners.Nodes, "No miners found: %v", strings.Join(output, "\n"))
+
+		// Use first miner
+		var miner climodel.Node
+		for _, m := range miners.Nodes[0] {
+			miner = m
+			break
+		}
+
+		// Get sharder list.
+		output, err = getSharders(t, configPath)
+		require.Nil(t, err, "get sharders failed", strings.Join(output, "\n"))
+		require.Greater(t, len(output), 1)
+		require.Equal(t, "MagicBlock Sharders", output[0])
+
+		var sharders map[string]climodel.Sharder
+		err = json.Unmarshal([]byte(strings.Join(output[1:], "")), &sharders)
+		require.Nil(t, err, "Error deserializing JSON string `%s`: %v", strings.Join(output[1:], "\n"), err)
+		require.NotEmpty(t, sharders, "No sharders found: %v", strings.Join(output[1:], "\n"))
+
+		// Use first sharder.
+		var sharder climodel.Sharder
+		for _, s := range sharders {
+			sharder = s
+			break
+		}
+
+		// Get base URL for API calls.
+		sharderBaseUrl := getNodeBaseURL(sharder.Host, sharder.Port)
+
+		// Get the starting balance for miner's delegate wallet.
+		res, err := apiGetBalance(sharderBaseUrl, miner.ID)
+		require.Nil(t, err, "Error retrieving miner %s balance", miner.ID)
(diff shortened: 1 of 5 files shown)


tests/cli_tests/block_rewards_test.go
```diff
@@ -0,0 +1,380 @@
...
+		err = json.Unmarshal([]byte(output[0]), &miners)
+		require.Nil(t, err, "Error deserializing JSON string `%s`: %v", output[0], err)
+		require.NotEmpty(t, miners.Nodes, "No miners found: %v", strings.Join(output, "\n"))
+
+		// Use first miner
+		var miner climodel.Node
+		for _, m := range miners.Nodes[0] {
+			miner = m
+			break
+		}
```
Kishan-Dhakan, line 67: I didn't understand this, can this not be just `miner := miners.Nodes[0]`?

tests/cli_tests/block_rewards_test.go
```diff
@@ -0,0 +1,380 @@
...
+		err = json.Unmarshal([]byte(strings.Join(output[1:], "")), &sharders)
+		require.Nil(t, err, "Error deserializing JSON string `%s`: %v", strings.Join(output[1:], "\n"), err)
+		require.NotEmpty(t, sharders, "No sharders found: %v", strings.Join(output[1:], "\n"))
+
+		// Use first sharder.
+		var sharder climodel.Sharder
+		for _, s := range sharders {
+			sharder = s
+			break
+		}
```
Kishan-Dhakan, line 85: Can this not be `sharder := sharders[0]` or am I missing something?

tests/cli_tests/block_rewards_test.go
```diff
@@ -0,0 +1,380 @@
...
+		require.Nil(t, err, "Error retrieving miner %s balance", miner.ID)
+		require.True(t, res.StatusCode >= 200 && res.StatusCode < 300, "Failed API request to check miner %s balance: %d", miner.ID, res.StatusCode)
+		require.NotNil(t, res.Body, "Balance API response must not be nil")
+
+		bt, err := ioutil.ReadAll(res.Body)
+		require.Nil(t, err, "Error reading response body")
+
+		var startBalance apimodel.Balance
+		err = json.Unmarshal(bt, &startBalance)
+		require.Nil(t, err, "Error deserializing JSON string `%s`: %v", string(bt), err)
```
Kishan-Dhakan, line 101: Can we make any assertions on Txn, Round and balance here other than the `Nil` on `err`?

tests/cli_tests/block_rewards_test.go
```diff
@@ -0,0 +1,380 @@
...
+		// Get base URL for API calls.
+		sharderBaseUrl := getNodeBaseURL(sharder.Host, sharder.Port)
+
+		// Get the starting balance for miner's delegate wallet.
+		res, err := apiGetBalance(sharderBaseUrl, miner.ID)
+		require.Nil(t, err, "Error retrieving miner %s balance", miner.ID)
+		require.True(t, res.StatusCode >= 200 && res.StatusCode < 300, "Failed API request to check miner %s balance: %d", miner.ID, res.StatusCode)
+		require.NotNil(t, res.Body, "Balance API response must not be nil")
+
+		bt, err := ioutil.ReadAll(res.Body)
```
Kishan-Dhakan, line 96: `bt` could be named better, what does bt stand for (I know it's the response body's content)?

tests/cli_tests/block_rewards_test.go
```diff
@@ -0,0 +1,380 @@
...
+			// otherwise, then remaining are distributed to stake holders.
+			if miner.TotalStake == 0 {
+				totalRewardsAndFees += int64(generatorRewardsRemaining)
+				totalRewardsAndFees += int64(generatorFeeRemaining)
+			}
+		}
+
+		wantBalanceDiff := totalRewardsAndFees
+		gotBalanceDiff := endBalance.Balance - startBalance.Balance
+		acceptableRoundingEpsilon := 100.0
```
Kishan-Dhakan, line 189: `InEpsilon` uses relative error to check if the actual and expected values are different. The formula used by them is `abs(actual - expected)/abs(actual)`, and if this is less than user given epsilon, the test passes ([here](https://github.com/stretchr/testify/blob/acba37e5db06f0093b465a7d47822bf13644b66c/assert/assertions.go#L1270)). Since both actual and expected are positive values, this will always be a number less than 1. The epsilon you have given is `100`, therefore, no matter the values of `wantBalanceDiff` & `gotBalanceDiff`. We need to change this epsilon value, you can use the `epsilon` variable (already declared in cli_tests package) which has a 0.1 value.

tests/cli_tests/block_rewards_test.go
```diff
@@ -0,0 +1,380 @@
...
+		assert.InEpsilonf(t, wantBalanceDiff, gotBalanceDiff, acceptableRoundingEpsilon, "expected total rewards and fees not matching: want %d, got %d", wantBalanceDiff, gotBalanceDiff)
+	})
+
+	t.Run("Sharder share on block fees and rewards", func(t *testing.T) {
```
Kishan-Dhakan, line 193: Same comments as for miner share case

tests/cli_tests/block_rewards_test.go
avanaur: For this one, given type returned is `map[string]climodel.Sharder` and key is node ID. The only way to get a node in the map is to do a for-loop.

```
./zwallet ls-sharders --all --json --silent
MagicBlock Sharders
{
  "4ab185573dbd84410ede7b6c6cbd53fe954a876168e3f40609f4bdcac51ae2d1": {
    "id": "4ab185573dbd84410ede7b6c6cbd53fe954a876168e3f40609f4bdcac51ae2d1",
    "version": "",
    "creation_date": 1633606385,
    "public_key": "fde3c3bc96b1b16e0545701e42583256859fa74a0c02894ef51e6492582ca00f9543ad2a714f1b817f46646354ed9c4fd2cb42e73785a4c29eeb7d6f635729a3",
    "n2n_host": "dev.0chain.net",
    "host": "dev.0chain.net",
    "port": 31102,
    "path": "sharder02",
    "type": 1,
    "description": "localhost.s02",
    "set_index": 0,
    "status": 0,
    "info": {
      "build_tag": "",
      "state_missing_nodes": 0,
      "miners_median_network_time": 0,
      "avg_block_txns": 0
    }
  },
  "fd30b6e59b2f467765dd03e5c9d9a51e4e5596eea63f17ffc1fa98fa9c635e5f": {
    "id": "fd30b6e59b2f467765dd03e5c9d9a51e4e5596eea63f17ffc1fa98fa9c635e5f",
    "version": "",
    "creation_date": 1633606385,
    "public_key": "5cf9588cfe20ea96e73935401f32a98190f325f673fda262f3e5d43d6d957609d4a6432d3e9a8437a4da36f0e6e4513b5b5f8aaae0fb6eecb69c22934ab0ba93",
    "n2n_host": "dev.0chain.net",
    "host": "dev.0chain.net",
    "port": 31101,
    "path": "sharder01",
    "type": 1,
    "description": "localhost.s01",
    "set_index": 1,
    "status": 0,
    "info": {
      "build_tag": "",
      "state_missing_nodes": 0,
      "miners_median_network_time": 0,
      "avg_block_txns": 0
    }
  }
}
```

avanaur:
> Tests are failing now. I run load test to dev earlier and it created some stakes. My expectation that `total_stake` will reflect the active stake of node must be incorrect. Have to investigate further where exactly is the issue.

Tests are all good now. PR should be good to review again

tests/cli_tests/block_rewards_test.go
```diff
@@ -0,0 +1,383 @@
...
+			require.Nil(t, err, "Error reading response body: %v", err)
+
+			var block apimodel.Block
+			err = json.Unmarshal(resBody, &block)
+			require.Nil(t, err, "Error deserializing JSON string `%s`: %v", string(resBody), err)
+
+			// No expected rewards for this miner if not the generator of block.
+			if block.Block.MinerId != miner.ID {
+				continue
+			}
```
Kishan-Dhakan, line 155: Not suggesting any changes, but out of curiosity, could it happen that in any round from `startBalance.Round+1` to `endBalance.Round`, the miner was never the generator of the block?

avanaur: yeah, there would be a chance. but very slim chance given we only have 3 miners. can improve it to repeatedly get endbalance until amounts are not equal

PR #82, opened 2021-10-20
Add file stats tests

littlestar642:
- file stats in root directory
- file stats in sub directory
- file stats in nested sub directory
- file stats on an empty allocation
- file stats for a file that does not exists
- file stats for an allocation you dont own
- file stats with no params supplied
- file stats with no remotepath param supplied
- file stats before and after download

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,7 +33,7 @@ linters-settings:
       - whyNoLint
       - wrapperFunc
   gocyclo:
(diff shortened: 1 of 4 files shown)


stewartie4:
hey @littlestar642 do you mind not force pushing to the branch? it destroys the branch history and makes it difficult to see what has changed in between reviews 
thanks

PR #87, opened 2021-10-24
WIP: ZCNSC tests (in progress)

Dmdv:
**ZCNSC system tests.**

These are imported tests and not ready for review yet.
Though some input is appreciated.

Some code will be removed and replaced with cli commands such as pouring and wallet registering.
Some zcnsc functions I will port to SDK


Dmdv:
> I don't see any tests here, just utility functions, can these be converted to use testing /go test like the rest of the repo? it would be better to keep consistency
> 
> Also is there a zcnsc cli? if so we should test via that rather than calling a hardcoded version of the go sdk

Sure thing :) 

I will remake all the code here.
I've just imported it.
I've added basic info in PR header to avoid misunderstanding )

Dmdv:
This branched is deprecated and will be replaced with 
https://github.com/0chain/system_test/pull/158

PR #99, opened 2021-11-03
Add or Remove Curator system tests

mohsenno1:


stewartie4 (approved):

tests/cli_tests/zboxcli_add_remove_curator_test.go
```diff
@@ -0,0 +1,388 @@
...
+		require.Nil(t, err, "could not get allocation ID", err, strings.Join(output, "\n"))
+		defer createAllocationTestTeardown(t, allocationID)
+
+		params := createParams(map[string]interface{}{"allocation": allocationID})
+		output, err = addCurator(t, params)
+		require.NotNil(t, err, "unexpected success on adding curator", strings.Join(output, "\n"))
+		require.Equal(t, "Error: curator flag is missing", output[0], strings.Join(output, "\n"))
+	})
+
+	t.Run("Add Curator _ Curator must be able to transfer the allocation ownership", func(t *testing.T) {
```
stewartie4, line 183: very minor: but I'd move the add tests before remove

tests/cli_tests/zboxcli_add_remove_curator_test.go
```diff
@@ -0,0 +1,388 @@
...
+		wallet, err := getWallet(t, configPath)
+		require.Nil(t, err, "Error occurred when retrieving wallet")
+
+		params := createParams(map[string]interface{}{"allocation": "INVALID ALLOCATION ID", "curator": wallet.ClientID})
+		output, err = addCurator(t, params)
+		require.NotNil(t, err, "unexpected success on adding curator", strings.Join(output, "\n"))
+		require.Contains(t, output[0], "adding curator:[txn] too less sharders to confirm it", strings.Join(output, "\n"))
+	})
+
+	t.Run("Add Curator _ must fail when the user is not the owner ", func(t *testing.T) {
```
stewartie4, line 132: This seems wrong to me
There is no such requirement that the curator must be the owner in the official docs.
Also if the only valid curator is the owner, why is there a "list" of curators (plural) 
The owner being the only valid curator defeats the purpose of the feature

tests/cli_tests/zboxcli_add_remove_curator_test.go
```diff
@@ -0,0 +1,388 @@
...
+		require.Equal(t, curatorWallet.ClientID, allocation.Curators[0], "Curator must've added to the allocation curators list")
+
+		output, err = transferAllocationOwnershipWithWallet(t, allocationID, curatorWalletName, targetWallet.ClientID, targetWallet.ClientPublicKey)
+		require.Nil(t, err, "error in transferring allocation as curator", strings.Join(output, "\n"))
+		require.Equal(t, 1, len(output), "unexpected output length:", strings.Join(output, "\n"))
+		expectedOutput := fmt.Sprintf("transferred ownership of allocation %s to %s", allocationID, targetWallet.ClientID)
+		require.Equal(t, expectedOutput, output[0], "unexpected output:", strings.Join(output, "\n"))
+	})
+
+	t.Run("Add Curator _ Owner added as curator must be able to transfer the ownership", func(t *testing.T) {
```
stewartie4, line 231: This test is the same as above, worth keeping if we change the test above to not be the owner

tests/cli_tests/zboxcli_add_remove_curator_test.go
```diff
@@ -0,0 +1,388 @@
...
+		require.Equal(t, "Error: curator flag is missing", output[0], strings.Join(output, "\n"))
+	})
+
+	t.Run("Add Curator _ Curator must be able to transfer the allocation ownership", func(t *testing.T) {
+		t.Parallel()
+
+		curatorWalletName := escapedTestName(t) + "_CURATOR"
+		targetWalletName := escapedTestName(t) + "_TARGET"
+
+		output, err := registerWallet(t, configPath)
```
stewartie4, line 189: We don't use this for anything in this test

tests/cli_tests/zboxcli_add_remove_curator_test.go
stewartie4, line 132: "Add Curator _ Curator must be able to transfer the allocation ownership" contradicts this test as a non-owner is added as a curator there

stewartie4 (changes requested):
some queries

tests/cli_tests/zboxcli_add_remove_curator_test.go
mohsenno1: No, this seems to be like bad naming. Here by "user" I mean the operator which is adding the curator.
So if you are not the owner of an allocation then you cannot add someone else as the curator to that allocation.

mohsenno1: I changed the test name to:

> Add Curator _ attempt to add curator by anyone except allocation owner must fail

tests/cli_tests/zboxcli_add_remove_curator_test.go
mohsenno1: The test are different. As I mentioned on the other comment, that is about testing if any one except allocation owner can add curator.
this is testing if after owner adds itself as curator is able to transfer the allocation.

tests/cli_tests/zboxcli_add_remove_curator_test.go
mohsenno1: you mean the wallet?
In this test we register a wallet and we create allocation by that. so this wallet is the owner. Then as the owner we add curatorWallet as curator. Then we test if curatorWallet is able to transfer allocation to a third party which is the targetWallet here.

Issue #32, opened 2021-09-17
System tests review and tidy up


stewartie4:
- consider better regex for USD -  [0-9]*[.])?[0-9]+ https://github.com/0chain/system_test/pull/28#discussion_r711050338
- add strings.join(output "\n") to require.Equal tests or change to require.Len(t, output, 0) 
- strip out blank and duplicate lines from command output
- remove boilerplate code from tests https://0chain.slack.com/archives/C02AV6MKT36/p1635274843078900?thread_ts=1635272798.078200&cid=C02AV6MKT36
- remove duplicated utility methods from tests https://0chain.slack.com/archives/C02AV6MKT36/p1635272798078200
- add additional calls to verify operations rather than just relying on cli exit status and message eg. list files after an upload, etc.
- add polling instead of explicit waits

Kishan-Dhakan:
Closing this as a stale issue.

PR #101, opened 2021-11-08
Miner fees payment

Kishan-Dhakan:
Added:
- Miner fee payment when using `rp-lock` with fees
- rp-unlock with fee
- sp-lock with fee
- sp-unlock with fee
- wp-lock with fee
- wp-unlock with fee
- zwallet lock and unlock command with fee


tests/cli_tests/miner_fees_payment_test.go
```diff
@@ -0,0 +1,868 @@
...
+		t.Parallel()
+
+		output, err := registerWallet(t, configPath)
+		require.Nil(t, err, "registering wallet failed", strings.Join(output, "\n"))
+
+		wallet, err := getWallet(t, configPath)
+		require.Nil(t, err, "Error occurred when retrieving target wallet")
+
+		output, err = executeFaucetWithTokens(t, configPath, 7)
+		require.Nil(t, err, "faucet execution failed", strings.Join(output, "\n"))
```
littlestar642, line 29: Can these three functions be grouped in a common function since they are being called at multiple places and should help in keeping the code simple?

tests/cli_tests/miner_fees_payment_test.go
```diff
@@ -0,0 +1,868 @@
...
+
+		mconfig := getMinerSCConfiguration(t)
```
littlestar642, line 31: golang has a convention of not adding `get` to getter functions. I m not sure but do we have different project level convention since I have seen a lot of getters being named like that 😅

tests/cli_tests/miner_fees_payment_test.go
```diff
@@ -0,0 +1,868 @@
...
+
+		// Create an allocation to use rp-lock on
+		allocParams := createParams(map[string]interface{}{
+			"expire": "5m",
+			"size":   "1024",
+			"lock":   "0.5",
+		})
+		output, err = createNewAllocation(t, configPath, allocParams)
+		require.Nil(t, err, "Failed to create new allocation", strings.Join(output, "\n"))
+
```
littlestar642, line 46: redundant line space

tests/cli_tests/miner_fees_payment_test.go
```diff
@@ -0,0 +1,868 @@
...
+						minerFeeRemaining := minerFee - minerServiceCharge
+
+						// If there is no stake, the miner gets entire fee.
+						// Else "Remaining" portion would be distributed to stake holders
+						// And hence not go the miner
+						if miner.TotalStake == 0 {
+							expectedMinerFee += minerFeeRemaining
+						}
+						t.Logf("Expected miner fee: %v", expectedMinerFee)
+						t.Logf("Miner ID: %v", block_miner_id)
```
littlestar642, line 115: can these debug logs be removed now?

tests/cli_tests/miner_fees_payment_test.go
```diff
@@ -0,0 +1,868 @@
...
+					if strings.ContainsAny(txn.TransactionData, "payFees") && strings.ContainsAny(txn.TransactionData, fmt.Sprintf("%d", transactionRound)) {
+						var transfers []apimodel.Transfer
+						err = json.Unmarshal([]byte(fmt.Sprintf("[%s]", strings.Replace(txn.TransactionOutput, "}{", "},{", -1))), &transfers)
+						require.Nil(t, err, "Cannot unmarshal the transfers from transaction output")
+
+						for _, transfer := range transfers {
+							// Transfer needs to be from Miner Smart contract to Generator miner
+							if transfer.From == MINER_SC_ADDRESS && transfer.To == block_miner_id {
+								t.Logf("--- FOUND IN ROUND: %d ---", block.Block.Round)
+								require.NotNil(t, transfer, "The transfer of fee to miner could not be found")
```
littlestar642, line 342: This check seems redundant given we enter the if block only when `transfer.From` and `transfer.To` has some value.

tests/cli_tests/miner_fees_payment_test.go
```diff
@@ -0,0 +1,868 @@
...
+					break
+				}
+			}
+			if found {
+				break
+			}
+		}
+		require.True(t, found, "Test Failed due to transfer from MinerSC to generator miner not found")
+
+		<-lockTimer.C
```
littlestar642, line 364: I did not get the use of lockTimer here. Can you please help?

tests/cli_tests/miner_fees_payment_test.go
```diff
@@ -0,0 +1,868 @@
...
+				if found {
+					break
+				}
+			}
+			if found {
+				break
+			}
+		}
+		require.True(t, found, "Test Failed due to transfer from MinerSC to generator miner not found")
+		wait(t, time.Minute)
```
avanaur, line 149: Would be best to be same as other tests that wait only until the lock has timed out

```
<-lockTimer.C
```

tests/cli_tests/miner_fees_payment_test.go
```diff
@@ -0,0 +1,868 @@
...
+							// Transfer needs to be from Miner Smart contract to Generator miner
+							if transfer.From == MINER_SC_ADDRESS && transfer.To == block_miner_id {
+								t.Logf("--- FOUND IN ROUND: %d ---", block.Block.Round)
+								require.NotNil(t, transfer, "The transfer of fee to miner could not be found")
+								// Transfer fee must be equal to miner fee
+								t.Log("Actual fee transfer: ", transfer.Amount, "Expected fee transfer:", expectedMinerFee)
+								require.InEpsilon(t, expectedMinerFee, transfer.Amount, 0.00000001, "Transfer fee must be equal to miner fee")
+								found = true
+							}
+							if found {
```
avanaur, line 134: can reduce the `if-founc-break` statements by adding a label to the outermost for-loop and add `break <label>` right after `found = true`.

https://www.ardanlabs.com/blog/2013/11/label-breaks-in-go.html

avanaur (approved):
LGTM. Although few suggestions from me

tests/cli_tests/miner_fees_payment_test.go
Kishan-Dhakan: Nice catch, I'll add this to #32

tests/cli_tests/miner_fees_payment_test.go
Kishan-Dhakan: These only show if the test fails, where they can help with debugging

stewartie4 (approved):

tests/cli_tests/miner_fees_payment_test.go
Kishan-Dhakan: No, this isn't redundant as there are plenty more transfers happening.

PR #104, opened 2021-11-11
fix(download): fixed assertions for download encrypted file

cnlangzi:


stewartie4:
cool, this bug has been fixed then 
thanks @cnlangzi

PR #106, opened 2021-11-11
Transfer Allocation tests

avanaur:
Cases covered:
- transfer allocation by owner should fail
- transfer allocation by curator
- transfer allocation by non-owner and non-curator should fail
- transfer allocation to self
- transfer an expired allocation
- transfer a canceled allocation (test is skipped - not doable)
- transfer a finalized allocation (test is skipped - not working)
- transfer allocation and download non-encrypted file
- transfer allocation and download encrypted file should fail
- transfer allocation and download with auth ticket should fail
- transfer allocation and upload file
- transfer allocation and update allocation
- transfer allocation with no allocation param should fail
- transfer allocation with no new_owner_key param should fail
- transfer allocation with no new_owner param should fail
- transfer allocation with invalid allocation param should fail
- transfer allocation with invalid new_owner param
- transfer allocation with invalid new_owner_key param should fail
- transfer allocation accounting test


tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,999 @@
...
+			"new_owner":     newOwnerWallet.ClientID,
+		})
+		require.NotNil(t, err, strings.Join(output, "\n"))
+		require.Greater(t, len(output), 1, strings.Join(output, "\n"))
+		require.Equal(t, "Error adding curator:[txn] too less sharders to confirm it: min_confirmation is 50%, but got 0/2 sharders", output[0])
+	})
+
+	t.Run("transfer a finalized allocation", func(t *testing.T) {
+		// FIXME
+		t.Skip("Finalizing allocation is not working")
```
stewartie4, line 228: instead of skipping this test here can you add it to broken scenarios file?
That way it will continue to run on master

tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,999 @@
...
+		require.Len(t, output, 1)
+
+		// if not empty, the transfer of allocation contents has occurred on blobbers.
+		// there is only one content expected so once it is no longer empty, transfer is deemed complete.
+		if output[0] != "[]" {
+			return true
+		}
+
+		select {
+		case <-timeout:
```
stewartie4, line 992: could you replace the switch statement with a simpler if statement?

avanaur: this is actually a select statement which checks if timeout channel is available, if not then it just continue

tests/cli_tests/zboxcli_transfer_allocation_test.go
avanaur: Decided to retain and assert the error.
Also, for some reason, cancel allocation is working so had that enable now.

avanaur: Hmm this is kinda tricky. Cancel allocation works on dev but not on CI run. Might just make it work for CI.

stewartie4: yeah @Sriep and myself have been chatting about this
instant cancel should work I think

tests/cli_tests/zboxcli_transfer_allocation_test.go
stewartie4: yeah I was just wondering if it was possible to do it as an if statement instead but it's minor, nvm

stewartie4 (approved):

tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,1002 @@
...
+			"allocation": allocationID,
+			"curator":    ownerWallet.ClientID,
+		}))
+		require.Nil(t, err, strings.Join(output, "\n"))
+		require.Len(t, output, 1)
+		require.Equal(t, fmt.Sprintf("%s added %s as a curator to allocation %s", ownerWallet.ClientID, ownerWallet.ClientID, allocationID), output[0])
+
+		// FIXME this does not work at the moment. Also error is not correct.
+		output, err = cancelAllocation(t, configPath, allocationID)
+		require.NotNil(t, err, strings.Join(output, "\n"))
```
littlestar642, line 204: Since this is not working, should we move this to broken scenarios test file?

tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,1002 @@
...
+		remotePath := "/child/" + filename
+
+		output, err = uploadFile(t, configPath, map[string]interface{}{
+			"allocation": allocationID,
+			"remotepath": remotePath,
+			"localpath":  file,
+		})
+		require.Nil(t, err, strings.Join(output, "\n"))
+		require.Len(t, output, 2)
+		require.Equal(t, "Status completed callback. Type = application/octet-stream. Name = "+filepath.Base(file), output[1])
```
Kishan-Dhakan, line 653: Is uploading a file necessary for this case?

tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,1002 @@
...
+		err = json.Unmarshal([]byte(output[0]), &finalWritePool)
+		require.Nil(t, err, "Error unmarshalling write pool info", strings.Join(output, "\n"))
+		require.Len(t, finalWritePool, 1)
+
+		actualCost := initialWritePool[0].Balance - finalWritePool[0].Balance
+
+		require.Equal(t, expectedTransferCost, actualCost)
+		require.True(t, finalWritePool[0].Locked, strings.Join(output, "\n"))
+		require.Equal(t, allocationID, finalWritePool[0].Id, strings.Join(output, "\n"))
+		require.Equal(t, allocationID, finalWritePool[0].AllocationId, strings.Join(output, "\n"))
```
Kishan-Dhakan, line 955: This might be what's happening right now but should it? I'd imagine if I'm no longer an owner of the allocation my tokens should be returned. New guy can pay for it, if new guys doesn't have enough wallet balance, transfer should fail.

thoughts @avanaur @stewartie4?

tests/cli_tests/zboxcli_transfer_allocation_test.go
Kishan-Dhakan: Ah I think it's needed for polling transfer complete got it

Kishan-Dhakan (approved):

tests/cli_tests/zboxcli_transfer_allocation_test.go
stewartie4: Agreed can you check up on this with the blockchain guys @avanaur ? if you don't get anywhere can you add a FIXME comment?

tests/cli_tests/zboxcli_transfer_allocation_test.go
stewartie4: The assertion is asserting on it not working so i think this is fine as-is so long as there is a FIXME comment (which there is)

tests/cli_tests/zboxcli_transfer_allocation_test.go
avanaur: Yeah, it is a hack. Can't think of another way to confirm transfer is complete. 
With a dummy file and list-all, I get a better confirmation that transfer is done given list-all requires a high consensus before it adds a file in the result.

tests/cli_tests/zboxcli_transfer_allocation_test.go
avanaur: Yeah, I'll add a FIXME and ask on the blockchain channel.

tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,999 @@
...
+			"allocation":    "badallocationid",
+			"new_owner_key": newOwnerWallet.ClientPublicKey,
+			"new_owner":     newOwnerWallet.ClientID,
+		})
+		require.NotNil(t, err, strings.Join(output, "\n"))
+		require.Greater(t, len(output), 1, strings.Join(output, "\n"))
+		require.Equal(t, "Error adding curator:[txn] too less sharders to confirm it: min_confirmation is 50%, but got 0/2 sharders", output[0])
+	})
+
+	// FIXME is this expected to fail?
```
mohsenno1, line 781: I'm agree with you, this should fail. Actually there is no check on curator or collaborator too. Literally you can add any string with any length as curator or collaborator. Here is the same. At least there should be a check to test if the parameter is valid node id. Or more advanced to check if the node exists or not.

tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,999 @@
...
+		// balance of new owner should be unchanged
+		output, err = getBalanceForWallet(configPath, newOwner)
+		require.Nil(t, err, "Unexpected balance check failure for wallet", escapedTestName(t), strings.Join(output, "\n"))
+		require.Len(t, output, 1)
+		require.Regexp(t, regexp.MustCompile(`Balance: 1.000 ZCN \(\d*\.?\d+ USD\)$`), output[0])
+
+		// zero cost to transfer
+		expectedTransferCost := int64(0)
+
+		// write lock pool of old owner should remain locked
```
mohsenno1, line 941:
```suggestion
		// write pool lock of old owner should remain locked
```

mohsenno1 (changes requested):

tests/cli_tests/zboxcli_create_dir_test.go
```diff
@@ -412,8 +412,12 @@ func createDirForWallet(cliConfigFilename, wallet string, withAllocationFlag boo
 }
 
 func listAll(t *testing.T, cliConfigFilename, allocationID string) ([]string, error) {
+	return listAllWithWallet(t, escapedTestName(t), cliConfigFilename, allocationID)
+}
+
+func listAllWithWallet(t *testing.T, wallet, cliConfigFilename, allocationID string) ([]string, error) {
 	time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
avanaur: I am not really keen to remove it on this PR given its another test entirely. 
We can look intor removing those explicit wait on another ticket and see if no issue with existing tests.
Thoughts @stewartie4 , @Kishan-Dhakan ?

tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,999 @@
...
+		output, err = transferAllocationOwnership(t, map[string]interface{}{
+			"allocation":    allocationID,
+			"new_owner_key": newOwnerWallet.ClientPublicKey,
+			"new_owner":     newOwnerWallet.ClientID,
+		})
+		require.Nil(t, err, strings.Join(output, "\n"))
+		require.Len(t, output, 1)
+		require.Equal(t, fmt.Sprintf("transferred ownership of allocation %s to %s", allocationID, newOwnerWallet.ClientID), output[0])
+
+		transferred := pollForAllocationTransferToEffect(t, newOwner, allocationID)
```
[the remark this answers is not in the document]
avanaur: Yes AFAIK that is correct, zbox waits for txn confirmation. However from testing the txn confirmation is not enough.

Upload, download, and update allocation fails most of the time when trying with the new owner right away. 

To have high confidence that those will work, need to wait until the new owner can list the files of the allocation.
Perhaps there is better approach but this is the only one that worked and make the test stable.

tests/cli_tests/zboxcli_transfer_allocation_test.go
avanaur: Yeah agreed. Should at least have check
- new_owner_key is a valid pubkey (length + char allower)
- new_ower is a valid client (length + char allowed)
- new_owner and new_owner_key are valid pairs

tests/cli_tests/zboxcli_transfer_allocation_test.go
```diff
@@ -0,0 +1,999 @@
...
+		downloadFilePath := strings.TrimSuffix(os.TempDir(), "/") + "/"
+		os.Remove(downloadFilePath + "/" + filename)
+
+		output, err = downloadFileForWallet(t, newOwner, configPath, createParams(map[string]interface{}{
+			"allocation": allocationID,
+			"localpath":  downloadFilePath,
+			"remotepath": remotePath,
+		}))
+		require.Nil(t, err, "Error in file operation", strings.Join(output, "\n"))
+		require.Len(t, output, 2)
```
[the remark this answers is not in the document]
avanaur: updated all to add extra message on assertions

tests/cli_tests/zboxcli_create_dir_test.go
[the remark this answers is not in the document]
stewartie4: I agree, I will add a point to https://github.com/0chain/system_test/issues/32

stewartie4:
I'll raise bugs for all of these

PR #110, opened 2021-11-17
Collaborator system tests

mohsenno1:


avanaur (commented):
Looks good to me. Covers a lot of tests already.

I am thinking if we should consider the following scenarios.
- are collaborators able to download encrypted file without issue?
- can a collaborator be added to a full folder and able to download contents?
- above and can a collaborator upload to folder?
- can we add success test to rename, delete, move, copy update attributes of file for added collaborator?
- can a collaborator remove  or add other collaborators? say to same remote path or to subfolder or contents?
- does collaborator pay for downloads for will respect the file's `who must pay` attribute? can we test for both owner and 3rd party attribute?

mohsenno1:
@avanaur 

> * are collaborators able to download encrypted file without issue?
> * can a collaborator be added to a full folder and able to download contents?
> * above and can a collaborator upload to folder?

I added two more cases:

- "collaborator can NOT be added to a directory"
- "collaborator should NOT be able to download encrypted file"

> * can a collaborator remove  or add other collaborators? say to same remote path or to subfolder or contents?

These cases are already covered:

> * can we add success test to rename, delete, move, copy update attributes of file for added collaborator?

This looks reasonable that a collaborator be able to update or edit the file, but this is not supported by the chain, at least not currently. We discussed with the team and it seems there is no active plan for adding these functionalities for now.

> * does collaborator pay for downloads for will respect the file's `who must pay` attribute? can we test for both owner and 3rd party attribute?

I think this one is more related to token-accounting test.

stewartie4:
@mohsenno1 
re: "can we add success test to rename, delete, move, copy update attributes of file for added collaborator?"
"This looks reasonable that a collaborator be able to update or edit the file, but this is not supported by the chain, at least not currently. We discussed with the team and it seems there is no active plan for adding these functionalities for now."

If that feature is not supported we need to add a test asserting which doesn't work 

"does collaborator pay for downloads for will respect the file's who must pay attribute? can we test for both owner and 3rd party attribute?"
"I think this one is more related to token-accounting test."
It would still be a good test case to have, can you add it?

mohsenno1:
> If that feature is not supported we need to add a test asserting which doesn't work

@stewartie4 The tests are already added:
Add Collaborator _ collaborator should NOT be able to update the file attribute
Add Collaborator _ collaborator should NOT be able to rename the file
Add Collaborator _ collaborator should NOT be able to delete the file
Add Collaborator _ collaborator should NOT be able to move the file
Add Collaborator _ collaborator should NOT be able to update the file
Add Collaborator _ collaborator should NOT be able to copy the file

> It would still be a good test case to have, can you add it?

I'll add this

stewartie4:
perfect! sorry i read the original comment wrong

stewartie4 (approved):
LGTM after that final test is added

PR #112, opened 2021-11-19
Share using auth ticket tests + improvements to existing share tests 

avanaur:
New
- Share unencrypted file to public using auth ticket
- Shared encrypted file to public using auth ticket should failed to download
- Revoke auth ticket on publicly-shared unencrypted file should fail
- Expired auth ticket of a publicly-shared file should fail to download
- Share to public a folder with no encrypted file using auth ticket
- Share encrypted huge file using auth ticket - proxy re-encryption (not working so was testing for failure)
- Share non-existent file should fail
- Share someone else's allocation file should fail
- Share file with missing allocation should fail
- Share file with missing remotepath should fail
- Share encrypted file using auth ticket - download accounting test - proxy re-encryption (not reducing the read pool atm)
- Share unencrypted file using auth ticket - download accounting test (not reducing the read pool atm)
- Share encrypted file using auth ticket - download accounting test where 3rd party pays - proxy re-encryption
- Share unencrypted file using auth ticket - download accounting test where 3rd party pays - proxy re-encryption 

Existing and renamed
- Share encrypted file using auth ticket - proxy re-encryption
- Revoke auth ticket of encrypted file - proxy re-encryption
- Expired auth ticket of an encrypted file should fail to download - proxy re-encryption
- Auth ticket for wrong clientId should fail to download - proxy re-encryption
- Auth ticket for wrong encryption public key should fail to download - proxy re-encryption
- Share folder with encrypted file using auth ticket - proxy re-encryption
- Folder not shared should fail to download - proxy re-encryption

Also include minor fixes
- Reduce balance locked on new alloc for test `Each blobber's read pool balance should reduce by download cost`
- Update `generateRandomTestFileName()` to ensure `/` is not doubled

CI Run: https://github.com/0chain/system_test/runs/4269095366?check_suite_focus=true


tests/cli_tests/flaky___broken_scenarios_test.go
```diff
@@ -177,7 +177,7 @@ func Test___FlakyBrokenScenarios(t *testing.T) {
 		require.Nil(t, err, "Failed to execute faucet transaction", strings.Join(output, "\n"))
 
 		allocParam := createParams(map[string]interface{}{
-			"lock":   balance,
+			"lock":   0.6,
```
stewartie4, line 180: is this no longer flaky after this change?

avanaur: works for a few run. can't say if no longer flaky. but I think looks good now

stewartie4: cool, we can keep it here and keep an eye on it
what does changing this from the balance variable to 0.6 do?

avanaur: it was locking 0.8 for allocation and 0.4 for read pool so going over 1 token. Fix is to just set the allocation to just 0.6

stewartie4 (approved):

Issue #114, opened 2021-11-20
free storage tests


stewartie4:
Investigate and Implement system acceptance tests that test the free storage functionality
- max number of free storage tokens per year
  
Following the[ test style guide](https://docs.google.com/document/d/12RtuT89FG7PycsOno7312yQoJ4nzzfs0xLBEDA38Q5k/edit?usp=sharing) and the style of the existing tests

moldis:
@stewartie4 Open question here, as we need to deploy chain with owners wallet.

stewartie4:
@moldis we should be able to hardcode the owner - @avanaur has done similar for the sc modification tests

avanaur:
https://github.com/0chain/zboxcli#free-storage-allocation

PR #125, opened 2021-12-06
Free storage tokens test

avanaur:
Resolves https://github.com/0chain/system_test/issues/114

Tests covered
- Create free storage from marker with accounting (currently not working)
- Create free storage with malformed marker should fail
- Create free storage with invalid marker contents should fail
- Create free storage with invalid marker signature should fail
- Create free storage with wrong recipient wallet should fail
- Create free storage with tokens exceeding assigner's individual limit should fail


tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
```diff
@@ -0,0 +1,474 @@
...
+	freeAllocAssignerTxn := freeAllocationAssignerTxn(t, ownerWallet, assignerWallet)
+	err = sendTxn(miners, freeAllocAssignerTxn)
+	require.Nil(t, err, "Error sending txn to miners: %v", output[0], err)
+
+	output, err = verifyTransaction(t, configPath, freeAllocAssignerTxn.Hash)
+	require.Nil(t, err, "Could not verify commit transaction", strings.Join(output, "\n"))
+	require.Len(t, output, 1)
+	require.Equal(t, "Transaction verification success", output[0])
+
+	// FIXME not working at the moment
```
mohsenno1, line 153: if the test is broken, shouldn't it be moved to `flaky___broken_scenarios_test.go`?

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
```diff
@@ -0,0 +1,474 @@
...
+	newValues := strings.Join([]string{
+		"1",    // decreasing data shards from default 10
+		"1",    // decreasing parity shards from default 5
+		"1024", // decreasing size from default 10000000000
+		"3m",   // increasing mcct from default 1m0s
+		"1",    // increasing read price range max from default 0.04
+		"1",    // increasing write price range max from default 0.04
+	}, ",")
+
+	output, err = updateStorageSCConfig(t, scOwnerWallet, map[string]interface{}{
```
mohsenno1, line 125: updating SC config may affect the other tests which are running in parallel. I think we tried to not be dependent to the chain configuration in tests, but I also believe this may cause flakiness in some tests.

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
```diff
@@ -0,0 +1,474 @@
...
+	require.Nil(t, err, "get miners failed", strings.Join(output, "\n"))
+	require.Len(t, output, 1)
+
+	var miners climodel.NodeList
+	err = json.Unmarshal([]byte(output[0]), &miners)
+	require.Nil(t, err, "Error deserializing JSON string `%s`: %v", output[0], err)
+	require.NotEmpty(t, miners.Nodes, "No miners found: %v", strings.Join(output, "\n"))
+
+	freeAllocAssignerTxn := freeAllocationAssignerTxn(t, ownerWallet, assignerWallet)
+	err = sendTxn(miners, freeAllocAssignerTxn)
```
mohsenno1, line 145: I'm wondering why do we directly send the transaction to miners? What is stopping us from using the zbox cli to create free allocation? The same logic currently exist on zbox.

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
```diff
@@ -0,0 +1,474 @@
...
+		require.Nil(t, err, "Could not write file marker")
+
+		output, err = createNewAllocationWithoutRetry(t, configPath, createParams(map[string]interface{}{"free_storage": markerFile}))
+		require.NotNil(t, err, "Failed to create new allocation", strings.Join(output, "\n"))
+		require.Greater(t, len(output), 1)
+		require.Equal(t, "Error creating free allocation: [txn] too less sharders to confirm it: min_confirmation is 50%, but got 0/2 sharders", output[0])
+	})
+}
+
+func getZCNWallet(t *testing.T, file string) *climodel.WalletFile {
```
mohsenno1, line 374: The function name is ambiguous.  
```suggestion
func readWalletFile(t *testing.T, file string) *climodel.WalletFile {
```

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
avanaur: the SC changes here are specific to free allocation

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
avanaur: there is no zbox for setting allocation assigner. there are no plans yet to add this on zbox according to @Sriep. 
@stewartie4 is aware of this reason why we have to do the signing free alloc assigner txn and sending to miners

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
avanaur: @stewartie4 is okay before to keep it here as long as it is passing. Passing in a sense it matches the current behaviour. As long as commented with FIXME to describe this should not be the behaviour. happy to move it to flaky test if needed

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
avanaur: updated

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
stewartie4: That would have been fine previously, but now that system tests is being used by everyone and blocks PRs i'd rather we assert on the expected result and move to broken scenarios

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
stewartie4: yep happy enough with this for now

stewartie4 (approved):

PR #128, opened 2021-12-10
Feature/move old flaky tests back into circulation

stewartie4:


tests/cli_tests/zboxcli_file_stats_test.go
```diff
@@ -427,6 +427,96 @@ func TestFileStats(t *testing.T) {
...
+			require.Equal(t, float64(data.NumOfBlocks), math.Ceil(float64(data.Size)/float64(chunksize)))
+			require.Equal(t, int64(1), data.NumOfUpdates)
+			if data.WriteMarkerTxn == "" {
+				require.Equal(t, false, data.BlockchainAware)
+			} else {
+				require.Equal(t, true, data.BlockchainAware)
+			}
+
+			//FIXME: POSSIBLE BUG: key name and blobberID in value should be same but this is not consistent for every run and happening randomly
+			// require.Equal(t, blobberID, data.BlobberID, "key name and blobberID in value should be same")
```
Kishan-Dhakan, line 474: IIRC @littlestar642 made a PR for this which was merged, so should be fixed by now, we can remove this

Kishan-Dhakan: Yep, [here](https://github.com/0chain/gosdk/pull/273)'s it and I checked using gosdk 1.3.4, it's fixed 🎉

Kishan-Dhakan (approved):

tests/cli_tests/zboxcli_file_stats_test.go
stewartie4: Ooh thanks for pointing that out @Kishan-Dhakan 
@littlestar642 thanks for the PR fix. the blobberID variable no longer exists in this test though - where is it supposed to come from?

stewartie4: nvm figured it out

stewartie4: hmmm i'm getting this output 
image
which is what we seen before - are we sure this change is in the latest staging release? or have i misunderstood the code change that needs made?
@littlestar642 / @Kishan-Dhakan

littlestar642: The code change should have resolved the issue 😕 . 
I m not sure if it was in the latest release though.

littlestar642 (approved):

tests/cli_tests/zboxcli_file_stats_test.go
Kishan-Dhakan: @stewartie4 @littlestar642 yeah this wouldn't work on CI yet, my bad. But it is fixed though. The problem is staging zboxcli still uses gosdk v1.3.1, while @littlestar642's merge was done after that (I used v1.3.4, it works fine then). The way I checked it manually by updating go.mod is not feasible through the CI, so I guess this would have to be left commented out for now 😞

PR #138, opened 2021-12-20
add check on length before assignment

Kishan-Dhakan:
It seems `owner_id` is returned as a single value instead of a pair sometimes,
image
which is causing out of bounds `panic` and [blocking PRs](https://github.com/0chain/blobber/runs/4583896134?check_suite_focus=true) for blobber team. Can we get this quick-fix merged?

### DIFF
--- a/tests/cli_tests/config/zbox_config.yaml
+++ b/tests/cli_tests/config/zbox_config.yaml
@@ -1,4 +1,4 @@
-block_worker: https://dev-3.devnet-0chain.net/dns
+block_worker: https://dev.0chain.net/dns
 signature_scheme: bls0chain
 min_submit: 50 # in percentage
 min_confirmation: 50 # in percentage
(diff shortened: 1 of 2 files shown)


tests/cli_tests/config/zbox_config.yaml
```diff
@@ -1,4 +1,4 @@
-block_worker: https://dev-3.devnet-0chain.net/dns
+block_worker: https://dev.0chain.net/dns
```
cnlangzi, line 1: it should be `https://dev-3.devnet-0chain.net/dns`

cnlangzi (changes requested):

tests/cli_tests/config/zbox_config.yaml
Kishan-Dhakan: Why?

cnlangzi: dev.0chain.net seem is not deployed for system_test?

cnlangzi: it seems is ok. I see it is replaced on github action with
`sed "s,block_worker:.*,block_worker: https://${{ env.NETWORK_URL }}/dns," -i ./config/zbox_config.yaml`

cnlangzi (approved):
LGTM

stewartie4 (approved):

PR #154, opened 2022-01-01
Feat/download case enhancement

Kishan-Dhakan:
Added coverage for `startBlock` and `endBlock` flags with `./zbox download` cmd:
- Download with positive startblock and positive endblock should work
- startblock 0 should fail
- negative startblock should fail
- negative endblock should fail
- download with only startblock should work
- download with only endblock should work
- endblock greater than maximum should fail
- endblock less than startblock should fail

### DIFF
--- a/tests/cli_tests/zboxcli_download_test.go
+++ b/tests/cli_tests/zboxcli_download_test.go
@@ -743,6 +743,39 @@ func TestDownload(t *testing.T) {
 		require.NotEqual(t, output[0], "invalid startblock. Please input greater than or equal to 1")
 	})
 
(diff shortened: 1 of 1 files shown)


stewartie4:
Thanks for this Kishan!!
Does this cover the issues that you've found? https://github.com/0chain/system_test/projects/2#card-75426641

stewartie4 (approved):

Kishan-Dhakan:
> Thanks for this Kishan!! Does this cover the issues that you've found? https://github.com/0chain/system_test/projects/2#card-75426641

Yep, it has those covered 👍

PR #162, opened 2022-01-11
Stake pool changes.

Sriep:
This makes changes necessary for https://github.com/0chain/0chain/pull/854, https://github.com/0chain/gosdk/pull/355 andhttps://github.com/0chain/zboxcli/pull/155. 

Change StakePoolOfferInfo and StakePoolDelegatePoolInfo to reflect changes on 0chain.
As we are now keeping a summary stats of offers rather than an array of offers, the existing test of comparing offers to allocation settings is removed as no longer available.


Sriep:
Green test run https://github.com/0chain/system_test/actions/runs/1745374265

Sriep:
https://github.com/0chain/system_test/actions/runs/1746786632 second green tests after some refactoring

internal/cli/model/model.go
```diff
@@ -245,16 +237,14 @@ type StakePoolRewardsInfo struct {
...
-	Interests        int64  `json:"interests"`         // total for all time
-	Penalty          int64  `json:"penalty"`           // total for all time
-	PendingInterests int64  `json:"pending_interests"` // total for all time
+	ID         string `json:"id"`          // pool ID
+	Balance    int64  `json:"balance"`     // current balance
+	DelegateID string `json:"delegate_id"` // wallet
+	Rewards    int64  `json:"rewards"`     // total for all time
+	Penalty    int64  `json:"penalty"`     // total for all time
 	// Unstake > 0, then the pool wants to unstake. And the Unstake is maximal
 	// time it can't be unstaked.
```
Kishan-Dhakan, line 246: Can we update these comments to reflect the changes?

tests/cli_tests/zboxcli_common_user_functions_test.go
```diff
@@ -114,32 +112,6 @@ func TestCommonUserFunctions(t *testing.T) {
...
-			allocationOffer := climodel.StakePoolOfferInfo{}
-			for _, offer := range stakePool.Offers {
-				if offer.AllocationID == allocationID {
-					allocationOffer = *offer
-				}
-			}
-
-			t.Logf("Expected blobber id [%v] to lock [%v] but it actually locked [%v]", blobber_detail.BlobberID, int64(blobber_detail.Size*int64(blobber_detail.Terms.Write_price)), int64(allocationOffer.Lock))
-			assert.Equal(t, int64(sizeInGB(blobber_detail.Size)*float64(blobber_detail.Terms.Write_price)), int64(allocationOffer.Lock))
-		}
```
Kishan-Dhakan, line 142: I get it that this changed, but what happens now? Do the blobbers don't lock anything? If so, these 2 cases should be removed completely (as in that case they are only testing creating and updating allocation, something that is covered separately). If not, the new token accounting must be added so it can be tested.

Kishan-Dhakan (changes requested):

tests/cli_tests/zboxcli_common_user_functions_test.go
Sriep: The information this test relied upon is no longer available. The test worked because the amount locked for an allocation was available from two sources and so could be compared. However, this redundancy is not scalable and has been removed. 

https://github.com/0chain/0chain/pull/854 replaced a slice that could easily be tens of thousands of entries long with a single summary statistic. This is all 0chain needs for its internal processing. 

If it turns out that users require the offer information it can be got from the event database. https://github.com/0chain/0chain/issues/814 I am not aware of such a requirement but we can add a new Event database table to cover it if necessary.

Also https://github.com/0chain/system_test/issues/172. Might be best done after https://github.com/0chain/0chain/pull/684, as I think we need some control over which blobbers are used for allocations.

Kishan-Dhakan: I see. [0chain/0chain#814](https://github.com/0chain/0chain/pull/854) mentions `TotalOffers`, can that be somehow used to test against this? If not, these two cases can be removed completely (and covered later when offer information is available again), as they don't check anything.

Sriep: Total offers is the total for the blobber, not the allocation. So to control this value you would need to take account of the blobbers being used in parallel tests.

Kishan-Dhakan: Okay, I will add a commit to remove these completely then 👍. Can you please edit [0chain/0chain#814](https://github.com/0chain/0chain/issues/814) to mention adding the necessary system_tests once stakePool table is available, so it doesn't get missed out? cc: @stewartie4 hope this is okay?

Kishan-Dhakan (approved):

tests/cli_tests/zboxcli_common_user_functions_test.go
stewartie4: yes this is important coverage to have, we'll need to re-add tests once the new method is available

stewartie4:
@Sriep these tests never passed before the PR was merged and are now failing on master - can they be reverted or can the required code be merged to staging please as this will break everyone

PR #164, opened 2022-01-16
Mn user info tests

Kishan-Dhakan:
Cases:
**./zwallet mn-user-info**
- Getting MinerSC Stake pools of a wallet before and after locking against a miner should work
- Getting MinerSC Stake pools of a wallet before and after locking against a sharder should work
- Getting MinerSC pools info for a different client id than wallet owner should work

### DIFF
--- a/tests/cli_tests/zwalletcli_minersc_pool_info_test.go
+++ b/tests/cli_tests/zwalletcli_minersc_pool_info_test.go
@@ -24,6 +24,9 @@ func TestMinerSCUserPoolInfo(t *testing.T) {
 	minerNodeDelegateWallet, err := getWalletForName(t, configPath, minerNodeDelegateWalletName)
 	require.Nil(t, err, "error fetching wallet")
 
+	sharderNodeDelegateWallet, err := getWalletForName(t, configPath, sharderNodeDelegateWalletName)
+	require.Nil(t, err, "error fetching wallet")
+
 	t.Run("Getting MinerSC Stake pools of a wallet before and after locking against a miner should work", func(t *testing.T) {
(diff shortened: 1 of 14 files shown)


tests/cli_tests/zwalletcli_minersc_pool_info_test.go
```diff
@@ -0,0 +1,236 @@
...
+		require.Equal(t, float64(1), intToZCN(poolsInfo.Pools["miner"][minerNodeDelegateWallet.ClientID][0].Balance))
+
+		// teardown
+		_, err = minerOrSharderUnlock(t, configPath, createParams(map[string]interface{}{
+			"id":      minerNodeDelegateWallet.ClientID,
+			"pool_id": poolId,
+		}), true)
+		if err != nil {
+			t.Log("error unlocking tokens after test: ", t.Name())
+		}
```
Kishan-Dhakan, line 76: Reason to add these:
- Each node can have at max 10 delegates. Besides the three here, if there are more than 7 cases in upcoming miner/sharder stake system tests (likely), they will start failing due to this. Hence, unlocking post test.

Since test itself is for mn-user-info and not unlock, instead of asserting the teardown I have just logged the error to avoid failing the tests in case unlock fails (which would be tested separately). This is not the usual way we do this, so I can add requires here if needed, just let me know in the comments

stewartie4: nah that makes good sense to me - teardown isn't part of the test so shouldn't fail it

stewartie4 (approved):

PR #171, opened 2022-01-26
Doc local test

Sriep:
Add documentation for local test against developer barnches

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,39 @@ PS: Test suite execution will be slower when running locally vs the system tests
 Output will also be less clear vs the system tests pipeline.   
 Therefore, we recommend using an IDE such as [GoLand/Intellij IDEA](https://www.jetbrains.com/go/) to run/debug individual tests locally
 
+## Run individual tests against local 0chain network
+
+For developing new system tests for code still in developer branches, tests can be run against a locally running chain.
+Typically, for a 0chain change you will have a PR for several modules that need to work
+together. For example, `0chain`, `blobber`, `GoSDK`, `zboxclie` and `zwalletclie`.
+
+The first step requires setting up a running chain using the GitHub branches from the PRs.
+Use the instructions for building a [local chain 0chain](https://github.com/0chain/0chain#setup-network),
+[add a few blobbers](https://github.com/0chain/blobber#building-and-starting-the-nodes).
+Make sure you [stake the blobbers](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/README.md#order).
+
+For `zboxcli` and `zwalletcle` changes you need to first build the executable and copy into local
+system test directory. For example:
+```bash
+cd zboxcli
+make install
+cp zbox ../system_test/tests/cli_tests/zbox
+```
+
+Make sure you have the correct system test branch. Now you need to edit `system_test/tests/cli_tests/config/zbox_config.yaml`
+Edit the line `block_worker: https://dev.0chain.net/dns` to the appropriate setting for you, something like
+```yaml
+block_worker: http://192.168.1.100:9091
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,6 +74,39 @@ PS: Test suite execution will be slower when running locally vs the system tests
...
+For developing new system tests for code still in developer branches, tests can be run against a locally running chain.
+Typically, for a 0chain change you will have a PR for several modules that need to work
+together. For example, `0chain`, `blobber`, `GoSDK`, `zboxcli` and `zwalletcli`.
+
+The first step requires setting up a running chain using the GitHub branches from the PRs.
+Use the instructions for building a [local chain 0chain](https://github.com/0chain/0chain#setup-network),
+[add a few blobbers](https://github.com/0chain/blobber#building-and-starting-the-nodes).
+Make sure you [stake the blobbers](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/README.md#order).
+
+For `zboxcli` and `zwalletcle` changes you need to first build the executable and copy into local
```
ssardana08, line 88:
```suggestion
For `zboxcli` and `zwalletcli` changes you need to first build the executable and copy into local
```

ssardana08 (changes requested):

README.md
```diff
@@ -72,7 +72,40 @@ go test ./... -v
...
+make install
+cp zbox ../system_test/tests/cli_tests/zbox
+```
+
+Make sure you have the correct system test branch. Now you need to edit `system_test/tests/cli_tests/config/zbox_config.yaml`
+Edit the line `block_worker: https://dev.0chain.net/dns` to the appropriate setting for you, something like
+```yaml
+block_worker: http://192.168.1.100:9091
+```
+Finally, you need to add a `system_test/tests/cli_tests/config/sc_owner_wallet.json` file with the
```
stewartie4, line 101: We can't include this in the README as it's sensitive data

README.md
```diff
@@ -72,7 +72,40 @@ go test ./... -v
...
-Therefore, we recommend using an IDE such as [GoLand/Intellij IDEA](https://www.jetbrains.com/go/) to run/debug individual tests locally
+Therefore, we recommend using an IDE such as [GoLand](https://www.jetbrains.com/go/) to run/debug individual tests locally
+
+## Run individual tests against local 0chain network
+
+For developing new system tests for code still in developer branches, tests can be run against a locally running chain.
+Typically, for a 0chain change you will have a PR for several modules that need to work
+together. For example, `0chain`, `blobber`, `GoSDK`, `zboxcli` and `zwalletcli`.
+
+The first step requires setting up a running chain using the GitHub branches from the PRs.
```
stewartie4, line 83: I would also mention that you can run tests against a remote chain if you have already deployed elsewhere eg. dev.0chain.net

README.md
Sriep: Is it sensitive data? For any important 0chain,  smart contract owners will set the owner from the config https://github.com/0chain/0chain/pull/685. So I think this is only used for testing. 

Regardless developers need to know about this.

README.md
Sriep: I have not used dev.0chain.net and I am not familiar with the details. It might be best for someone familiar with the process to add a block about using dev.0cahin.net. I added I line just mentioning the possablity.

README.md
stewartie4: Yeah it is because the same keys are used in our public networks but this is a minor point - maybe we're better committing this to repo if it's just going to be in a readme

Sriep: I changed the wallet text to ` PM the system test team for a copy of the wallet.json file.`. Does this work?

ssardana08:
@Sriep @stewartie4 can we merge this?
it's extremely useful while debugging failures on local chain.

stewartie4:
Sure! I forgot all about this one 
Thanks for the contribution

PR #178, opened 2022-02-01
Pr merging protocol

Kishan-Dhakan:
Make additions to README.md


internal/cli/util/utils.go
```diff
@@ -51,6 +51,11 @@ func RunCommand(t *testing.T, commandString string, maxAttempts int, backoff tim
 	var count int
 	for {
 		count++
+
+		if count == maxAttempts {
+			commandString = strings.Replace(commandString, "--silent", "", 1)
```
stewartie4, line 56: good idea but this will mess with output assertions right?

Kishan-Dhakan: Yep. I changed it a bit to get us the logging without messing output assertions

stewartie4: slightly concerned that this will lead to very unwieldy log output in the console but we'll see LGTM for now

stewartie4 (approved):

stewartie4:
Thanks for the detailed README Kishan

PR #187, opened 2022-02-08
remove parallel

Kishan-Dhakan:
Removing parallel running of these cases because:
- They are very small tests 
image
- Since there is shared data (account based on same mnemonic) I suspect running in parallel is introducing flakiness (see [here](https://github.com/0chain/blobber/runs/5114094520?check_suite_focus=true) and [here](https://github.com/0chain/blobber/actions/runs/1810379214/attempts/1))


dabasov (approved):
lgtm

PR #192, opened 2022-02-11
fix unfinished share test

stewartie4:


tests/cli_tests/zboxcli_share_file_test.go
```diff
@@ -187,13 +187,11 @@ func TestShareFile(t *testing.T) {
...
 			"authticket": authTicket,
 		})
-		_, err = downloadFileForWallet(t, receiverWallet, configPath, downloadParams, false)
-		require.NotNil(t, err)
-		/* TODO: Enable output assertion
+		output, err = downloadFileForWallet(t, receiverWallet, configPath, downloadParams, false)
+		require.NotNil(t, err, "Expected error to be present but was not.", strings.Join(output, "\n"))
 		require.Len(t, output, 2, "download file - Unexpected output", strings.Join(output, "\n"))
-		require.Equal(t, "Status completed callback. Type = application/octet-stream. Name = "+filepath.Base(file), output[1],
+		require.Equal(t, "Error in file operation: File content didn't match with uploaded file",
```
stewartie4, line 193: this output doesn't seem correct  given the scenario @lpoli it suggests that the file has been downloaded but can't be verified?
compare with https://github.com/0chain/system_test/blob/master/tests/cli_tests/zboxcli_download_test.go#L1244 where we download without sharing

stewartie4: This test is now failing consistently on staging 
image
Error is not being returned as expected

lpoli: @stewartie4 Lets only check if error is nil or not for now.
The error response from download function is not proper. Please raise issue for this in gosdk and assign me.

The error it is getting from blobber is `Response Error: client does not have permission to download the file. share revoked`
But download function returns `Error in file operation: File content didn't match with uploaded file`

revoke

stewartie4: Hey @lpoli the issue is this is not always happening - sometimes the error is nil instead of not nil
image

Kishan-Dhakan, line 193: Can confirm, this is flaky

lpoli: @stewartie4 The code that checks shareInfo is:
https://github.com/0chain/blobber/blob/staging/code/go/0chain.net/blobbercore/handler/object_operation_handler.go#L280-L287

You can see that there's nothing to modify.

It might be that shared client is either owner or collaborator. In this case authticket is redundant, so never checked. In other case it should obviously fail.

stewartie4: I would suggest that this should either pass all the time or fail all the time 
If we're getting into a situation where identical test data is failing sometimes and passing other times then we have flaky/broken functionality 
Thoughts?

stewartie4:
@lpoli is looking into fixing the comment above

stewartie4:
@lpoli is looking into fixing the comment above

PR #215, opened 2022-03-20
System Tests Resiliency

Kishan-Dhakan:
## Changelog

- `TestMinerUpdateSettings` and `TestSharderUpdateSettings` were removed from running in parallel till [hotfix/staging-stabilization](https://github.com/0chain/0chain/tree/hotfix/staging-stabilization) was merged. It was merged so this has been reverted and tests are back to parallel.

- Retries were set to false for `TestFileDelete/delete_existing_file_with_thumbnail_should_work` which caused flakiness as shown below but passed on retry, resolved.

image

- All `TestMinerUpdateSettings` and `TestSharderUpdateSettings` test cases use the same wallet (owner's). Running these in parallel could cause race conditions which cause flakiness as shown below, resolved.

image

- Retries in `TestFileUpdate` were turned off, which caused flakiness as shown below, resolved.

image

- All thumbnail tests were using the same filename `upload_thumbnail_test.png` for thumbnail files. This caused race conditions and flakiness as shown below. Resolved by changing filename to depend on `t.Name()`.

image

- `CreateAllocationFreeStorage` have been flaky for quite some time as shown below, but there is another reason to move them `FlakyScenarios`. The only success scenario test case in this file is `FIXME` at the moment. As "unrelated failures" often confuse developers even if they pass on retries, and noting that this feature/test is not working at the moment anyway, I have moved it to `Test___FlakyScenariosCreateAllocationFreeStorage`.

image

- `"transfer allocation by owner should fail"` case was not updated after a PR changing this was merged, so I have updated it to `"transfer allocation by owner should work"`.

- `TestBlobberConfigUpdate` had a `defer func` call inside each case to revert the changes of previous test. This often caused weird flakiness as shown below. I have refactored those calls into a `t.Cleanup()` in the main test, which takes care of reverting without causing flakiness.

image

- `TestVestingPool` was a huge monolith that caused flakiness as shown below. I have broken it to 7 different test files, namely
 `zwalletcli_vesting_pool_unlock_test.go`, `zwalletcli_vesting_pool_test.go`, `zwalletcli_vesting_pool_stop_test.go`, `zwalletcli_vesting_pool_list_test.go`, `zwalletcli_vesting_pool_info_test.go`, `zwalletcli_vesting_pool_delete_test.go`, `zwalletcli_vesting_pool_add_test.go`.

image

- Broke `TestLockAndUnlockInterest` into `TestLockAndUnlockInterest` and `TestUnlockInterest`.

- Added missing checks like `NotEmpty()`, `Len()` across 40+ files to prevent panics (`Note: a panic thrown in the test suite fails all remaining tests immediately, so this is necessary`). Also refactored code like `require.Equal(t, len(arr), 5)` to `require.Len()`, removed redundant logging like `require.Nil(t, err, "msg", err)`, and other smaller maintenance tasks.

- Sample video for `Upload Video File Should Work` was run without retries which caused flakiness as shown below. Resolved.

image

## Result

As a result of above changes, flakiness has been reduced greatly. The *only* flaky test now is `TestBlobberChallenge` which passes on retries. One other case that fails and passes on retry is any case from `TestDelete` but this is not flakiness, is an underlying issue of handling partial failures discussed here:  https://0chain.slack.com/archives/GREHVKQKS/p1644797353994499.

See five consecutive runs did after these changes:
1. https://github.com/0chain/system_test/actions/runs/2011160446/attempts/1
2. https://github.com/0chain/system_test/actions/runs/2011160446/attempts/2
3. https://github.com/0chain/system_test/actions/runs/2011160446/attempts/3
4. https://github.com/0chain/system_test/actions/runs/2011160446/attempts/4
5. https://github.com/0chain/system_test/actions/runs/2011160446


stewartie4 (approved):
LGTM

stewartie4:
shall we merge this before there are too many merge conflicts @Kishan-Dhakan ? or are we waiting for other PRs to go first?

Kishan-Dhakan:
Yep, I'm thinking merging now would be great

Kishan-Dhakan:
@stewartie4 I'm good to go with merging

service-0chain:
Manual system tests [success] with the following config
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | feat/parallelize-config-tests  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #220, opened 2022-03-26
Feature/add config to repo

stewartie4:
actions PR: https://github.com/0chain/actions/pull/27

Todo: add ci code to delete wallet files when running against custom network as we can't guarantee these will work?
(thoughts on this approach would be great)


tests/cli_tests/zwalletcli_zcnbridge_ethereum-account-register_test.go
```diff
@@ -29,7 +29,7 @@ func TestEthRegisterAccount(t *testing.T) {
 
 	t.Run("Register ethereum account in local key storage", func(t *testing.T) {
 		deleteDefaultAccountInStorage(t, address)
-		output, err := importAccount(t, mnemonic, password, false)
+		output, err := importAccount(t, password, mnemonic, false)
```
stewartie4, line 32: The ordering of params is incorrect on master

internal/cli/util/utils.go
```diff
@@ -64,15 +64,16 @@ func RunCommand(t *testing.T, commandString string, maxAttempts int, backoff tim
...
 
-			if err != nil {
-				t.Logf("%sThe verbose output for the command is:", red)
-				commandString = strings.Replace(commandString, "--silent", "", 1)
-				out, _ := RunCommandWithoutRetry(commandString) // Only for logging!
-				for _, line := range out {
-					t.Logf("%s%s", red, line)
+			if strings.EqualFold(strings.TrimSpace(os.Getenv("DEBUG")), "true") {
+				if err != nil {
+					t.Logf("%sThe verbose output for the command is:", red)
```
stewartie4, line 69: putting this behind a debug flag because it makes test output very unwieldy when multiple tests fail, plus the information is already available in the archived logs

Kishan-Dhakan: IIRC in the archived logs it is difficult to know which logs belong to which transactions. I'd suggest keeping this as is, it makes diagnosing the failures a lot easier, even if there are too many of them.

tests/cli_tests/zwalletcli_zcnbridge_ethereum-account-register_test.go
Kishan-Dhakan: Good catch

.github/workflows/ci.yml
```diff
@@ -64,7 +64,7 @@ jobs:
 
   system-tests:
     name: "System Tests"
-    runs-on: [ tests-suite ]
+    runs-on: [ tests-suite05 ]
```
Kishan-Dhakan, line 67: Out of curiosity, what does this do?

dabasov: yep, the same question

stewartie4: this makes my branch only run on the tests-suite05 runner 
the reason being I don't want to affect other runners than this until I merge this 
it will be reverted back to tests-suite before I merge

internal/cli/util/utils.go
stewartie4: sure, I'll revert

stewartie4:
@Kishan-Dhakan @dabasov @cnlangzi ready for re-review, the last thing i have to do here is revert to use master actions branch and not a specific runner so this should be able to be approved in anticipation of that if you guys are happy

Kishan-Dhakan (approved):
LGTM

stewartie4:
tests passing here:
https://github.com/0chain/system_test/runs/5808783371?check_suite_focus=true

PR #253, opened 2022-04-30
Feature/initial api test framework

stewartie4:
Rough initial framework to create wallet + allocation, ready for additional API tests


internal/api/util/Config.go
```diff
@@ -0,0 +1,22 @@
+package util
+
+import (
+	"io/ioutil"
+
+	"gopkg.in/yaml.v3"
+)
```
stewartie4, line 7: this is clearly used on line 18

internal/api/util/Config.go
```diff
@@ -0,0 +1,22 @@
...
+	NetworkEntrypoint string `yaml:"network_entrypoint"`
+}
+
+func (config *Config) Init(path string) {
+	yamlFile, err := ioutil.ReadFile(path)
+	if err != nil {
+		panic("Failed to read config file!")
+	}
+	err = yaml.Unmarshal(yamlFile, config)
+	if err != nil {
```
stewartie4, line 19: see import

stewartie4: the rest of these are the same... i'm obviously doing something wrong

internal/api/model/Model.go
```diff
@@ -1,7 +1,11 @@
-package apimodel
+package model
```
Kishan-Dhakan, line 1: Why rename this file to titlecase?😅 Conventionally go files are snake case or all lower case names, although this is not documented strictly anywhere. Even so, to be consistent in this repo we could stick to one style

Kishan-Dhakan (approved):
LGTM

internal/api/model/Model.go
stewartie4: I'm used to java i guess lol i'll revert it

PR #299, opened 2022-06-13
Feature/remove internal retries

stewartie4:
Removes command retries from tests.
These "internal" retries were added due to chain flakiness and prevent a "blip" from failing the entire test.
However, this close to mainnet we should be investigating these "blips" instead of masking them.
Failing tests will still be retried for the sake of CI stability, but individual commands will no longer be retried "internally" to a test itself


stewartie4 (commented):
my gut tells me we're still too unstable for this, thoughts?

stewartie4 (commented):
Out of date

PR #331, opened 2022-07-15
remove zproxy

Kishan-Dhakan:
- Removes zproxy from network as it is deprecated.


stewartie4 (approved):
LGTM after revert to master

Kishan-Dhakan:
Green run: https://github.com/0chain/system_test/runs/7368892728?check_suite_focus=true, but it failed to set PR status to passed. Merging

PR #334, opened 2022-07-18
Free allocation tests

Kishan-Dhakan:


tests/cli_tests/config/zbox_config.yaml
```diff
@@ -1,4 +1,4 @@
-block_worker: https://dev-1.devnet-0chain.net/dns
+block_worker: https://dev-2.devnet-0chain.net/dns
```
stewartie4, line 1: minor: revert

stewartie4 (commented):
what's the status of this PR @Kishan-Dhakan ?

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
```diff
@@ -29,8 +30,261 @@ const (
...
+		require.Nil(t, err, "Could not write file marker")
+
+		output, err = createNewAllocationWithoutRetry(t, configPath, createParams(map[string]interface{}{"free_storage": markerFile}))
+		require.NotNil(t, err, "Failed to create new allocation", strings.Join(output, "\n"))
+		require.Len(t, output, 1)
+		require.Equal(t, "unmarshalling markerinvalid character 'b' looking for beginning of value", output[0])
+	})
+
+	t.Run("Create free storage with invalid marker contents should fail", func(t *testing.T) {
+		// register recipient wallet
```
stewartie4, line 161: can these tests run in parallel? if not we should probably prepend a zero to the test file to ensure they run at the beginning with the other sequential tests

stewartie4 (changes requested):
LGTM other than comment re: parallel tests

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
Kishan-Dhakan: @stewartie4 the internal tests can't be run in parallel as they show flakiness, but the entire can be run along with other files. There's a top level t.Parallel() for that.

stewartie4 (approved):

PR #343, opened 2022-07-27
Enable FauscetSC config update

sum-it08:


stewartie4 (approved):

Kishan-Dhakan:
IIRC the reason this was skipped was because `faucet` would be off for mainnet.

PR #370, opened 2022-08-09
Improve blobber tests

Kishan-Dhakan:
- Increase file size for large upload case to 1 GB
- Add rename root path coverage

### DIFF
--- a/tests/cli_tests/zboxcli_file_rename_test.go
+++ b/tests/cli_tests/zboxcli_file_rename_test.go
@@ -526,6 +526,28 @@ func TestFileRename(t *testing.T) { // nolint:gocyclo // team preference is to h
 		require.True(t, foundAtDest, "file not found at destination: ", strings.Join(output, "\n"))
 	})
 
+	t.Run("rename file", func(t *testing.T) {
+		t.Parallel()
+
+		allocSize := int64(2048)
+
+		remotePath := "/"
+		destName := "new_"
+
+		allocationID := setupAllocation(t, configPath, map[string]interface{}{
+			"size": allocSize,
+		})
+
+		output, err := renameFile(t, configPath, map[string]interface{}{
+			"allocation": allocationID,
+			"remotepath": remotePath,
+			"destname":   destName,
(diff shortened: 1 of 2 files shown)


tests/cli_tests/zboxcli_file_rename_test.go
```diff
@@ -526,6 +526,28 @@ func TestFileRename(t *testing.T) { // nolint:gocyclo // team preference is to h
 		require.True(t, foundAtDest, "file not found at destination: ", strings.Join(output, "\n"))
 	})
 
+	t.Run("rename file", func(t *testing.T) {
```
stewartie4, line 529: rename root path directory fail?

stewartie4 (changes requested):
minor test rename needed if I'm right

tests/cli_tests/zboxcli_upload_test.go
```diff
@@ -312,11 +312,13 @@ func TestUpload(t *testing.T) {
 	t.Run("Upload Large File Should Work", func(t *testing.T) {
 		t.Parallel()
 
-		allocSize := int64(500 * MB)
-		fileSize := int64(99 * MB)
+		allocSize := int64(2 * GB)
```
stewartie4, line 315: does 2gb and 1gb give us anything that 500 and 100mb didn't?

stewartie4 (changes requested):
Happy to make these changes @Kishan-Dhakan ?

tests/cli_tests/zboxcli_file_rename_test.go
Kishan-Dhakan: Missed that, Done

tests/cli_tests/zboxcli_upload_test.go
Kishan-Dhakan: @lpoli suggested this would be a better benchmark. See: https://0chain.slack.com/archives/C03788G8N2D/p1660054501357799

stewartie4 (approved):

PR #375, opened 2022-08-18
change rename error

Kishan-Dhakan:


lpoli (approved):
Most of the times it is good to `lowercase` both expected and output string, so it will be easier and less error prone.

Issue #361, opened 2022-07-27
fix skipped TestMinerSharderPoolInfo


sum-it08:
=== SKIP: tests/cli_tests TestMinerSharderPoolInfo (0.00s)

PR #378, opened 2022-08-22
Pool info

Kishan-Dhakan:
- Unskip and fix TestMinerSharderPoolInfo

Closes https://github.com/0chain/system_test/issues/361


tests/cli_tests/main_test.go
```diff
@@ -18,8 +18,14 @@ const (
 	miner03NodeDelegateWalletName   = "wallets/miner03_node_delegate"
 	sharder01NodeDelegateWalletName = "wallets/sharder01_node_delegate"
 	sharder02NodeDelegateWalletName = "wallets/sharder02_node_delegate"
-	minerNodeWalletName             = "wallets/miner_node"
-	sharderNodeWalletName           = "wallets/sharder_node"
+)
+
+const (
+	miner01ID   = "73ad5727612116c025bb4405bf3adb4a4a04867ae508c51cf885395bffc8a949"
```
stewartie4, line 24: why do we need to hardcode these ids? the tests should be able to run against any network and x number of miners/sharders

stewartie4 (changes requested):
Some minor questions

stewartie4:
So so great to get these working!

tests/cli_tests/zwalletcli_sharder_update_settings_test.go
```diff
@@ -250,7 +429,17 @@ func TestSharderUpdateSettings(t *testing.T) {
 	})
 
 	t.Run("Sharder update settings from non-delegate wallet should fail", func(t *testing.T) {
-		t.Parallel()
```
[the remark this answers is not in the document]
Kishan-Dhakan: All these settings update tests use the same `delegate wallet`. If run in parallel, nonce will mismatch and cause flakiness. So I'm afraid these will have to be run sequentially.

tests/cli_tests/main_test.go
Kishan-Dhakan: It's because we need to hardcode the delegate wallets. Earlier, the operational wallet (miner.ID == wallet.ClientID) was being used for this (because the operational wallet is the default delegate wallet), which meant we could get these dynamically from ls-miners. But the operational wallet is used for a lot of transactions (like payFees) in the background, so it causes nonce-mismatch, and also for security purposes we should not be keeping delegate and operational wallet to be the same. Since I have changed them, I need to hardcode these IDs against their respective wallets, there is no way to do this dynamically (which makes sense).

It might be worth noting that even earlier, we couldn't have run these against any network with X no. of M/S because we would still need to hardcode all their wallets as well.

stewartie4: hmmmm ok, can we pass these ids through as run params rather than hardcoding? 
Then if they're not passed in, these tests can be skipped 
It's just as it is now these tests will run and fail against every network other than dev-x

service-0chain:
Manual system tests [success] with the following config
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | poolInfo  |
| gosdk  | fix/cornerCaseConsensus  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
Manual system tests [cancelled] with the following config
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | poolInfo  |
| gosdk  | fix/cornerCaseConsensus  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

Issue #64, opened 2021-10-04
wallet concurrency tests


stewartie4:
https://github.com/0chain/blobber/wiki/Features-Upload#upload

PR #395, opened 2022-09-03
Feature/wallet concurrency tests

YarikRevich:
PR refers to issue #64

Added concurrent tests for all file requests


tests/cli_tests/zboxcli_file_rename_test.go
```diff
@@ -87,6 +89,158 @@ func TestFileRename(t *testing.T) { // nolint:gocyclo // team preference is to h
...
+				renameOutputList[currentIndex] = op
+
+				op, err = deleteFile(t, escapedTestName(t), createParams(map[string]interface{}{
+					"allocation": allocationID,
+					"remotepath": filepath.Join(remotePathPrefix, destFileName),
+				}), true)
+
+				deleteErrorList[currentIndex] = err
+				deleteOutputList[currentIndex] = op
+			}(i)
```
Kishan-Dhakan, line 209: This seems to do delete and rename twice. Would be better to upload two files, then `go func(){ rename }` and `go func() { delete }`.

Kishan-Dhakan (changes requested):
Minor comment: Great work otherwise.

tests/cli_tests/zboxcli_file_rename_test.go
YarikRevich: You know, that there won't be any guarantee, that file will be renamed, so it won't be possible to test it correctly

YarikRevich: The only way is to use mutex, but don't know, if it will make any difference

tests/cli_tests/zboxcli_file_rename_test.go
```diff
@@ -87,6 +89,178 @@ func TestFileRename(t *testing.T) { // nolint:gocyclo // team preference is to h
...
+				}), true)
+
+				deleteErrorList[currentIndex] = err
+				deleteOutputList[currentIndex] = op
+
+				mutex.Unlock()
+			}(i)
+		}
+
+		wg.Wait()
```
Kishan-Dhakan, line 232: What we're trying to test is that concurrent delete and rename operations work on the same `allocation`, not file. Can you please update the flow to:
- Upload two files
- go func() { rename file 1 }
- go func() { delete file 2 }

So it does not need to be on the same file as long as the it's the same allocation.

Kishan-Dhakan (changes requested):

tests/cli_tests/zboxcli_file_rename_test.go
YarikRevich: Got it, fixed

tests/cli_tests/zboxcli_file_rename_test.go
```diff
@@ -170,62 +170,61 @@ func TestFileRename(t *testing.T) { // nolint:gocyclo // team preference is to h
 			"size": allocSize,
 		})
 
-		var fileNames [2]string
+		var renameFileNames [2]string
 		var destFileNames [2]string
```
Kishan-Dhakan, line 174: Minor: Do we need to two for each?

YarikRevich: Yes, we need

PR #404, opened 2022-09-11
Fix: Skipped TestShareFile tests

nlok5923:
Fixing the skipped TestShareFile test.


tests/cli_tests/zboxcli_share_file_test.go
```diff
@@ -1142,7 +1141,7 @@ func TestShareFile(t *testing.T) {
 		require.Nil(t, err, "Error unmarshalling read pool", strings.Join(output, "\n"))
 		require.NotEmpty(t, initialReadPool)
 
-		require.Equal(t, 0.4, initialReadPool.Balance)
+		require.Equal(t, 0.4*1e10, float64(initialReadPool.Balance))
```
Kishan-Dhakan, line 1144: The readPool you are checking before and after downloading is of the owner. But the download is being done by `receiverWallet` here: `output, err = downloadFileForWallet(t, receiverWallet, configPath, downloadParams, false)`.

Kishan-Dhakan (changes requested):

tests/cli_tests/zboxcli_share_file_test.go
nlok5923: Fixed this in latest commit.

nlok5923:
@Kishan-Dhakan the test are passing on test run here https://github.com/0chain/system_test/actions/runs/3216728675/jobs/5258912965 on https://github.com/0chain/0chain/pull/1721 pr but failing due authorizer crashing 
image

stewartie4 (changes requested):
Few questions, why are these being skipped? Can you update the t.skip reason? 
Also why remove the logging?

nlok5923:
> Few questions, why are these being skipped? Can you update the t.skip reason? Also why remove the logging?

Hi @stewartie4 
> why are these being skipped? 

Actually the test are being skipped because on sharing file and downloading file by receiver the read pool balance is not affected although it incurred some download cost due to downloading.

> Can you update the t.skip reason?

As now the test i fixed i think we won't be keeping the `t.Skip` statement. so shall i still update the reason.

> Also why remove the logging?

I had added those statement just for debugging purpose.

Kishan-Dhakan:
Failing due to unrelated authorizer crashes, merging this.

PR #409, opened 2022-09-13
Feature/update blobber tests

YarikRevich:
Refers to issue #407


stewartie4 (commented):
Why was this closed? these changes look great @YarikRevich

PR #415, opened 2022-09-19
Feature/handle panic in crypto

stewartie4:


internal/api/util/crypto/crypto.go
```diff
@@ -7,7 +7,7 @@ import (
 	_ "crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"github.com/0chain/gosdk/core/encryption"
```
stewartie4, line 10: we shouldn't be using gosdk during API tests - the point of these tests is to test the contract outside of any changes we make to gosdk

peterlimg (approved):
lgtm 👍

Issue #408, opened 2022-09-13
Refactor util functions for API tests


YarikRevich:

stewartie4:
Hi @YarikRevich what is this issue trying to solve?

YarikRevich:
Right now test environment is not scalable and predictable. This issue tries to solve it

stewartie4:
It's better for issues raised to be specific and measurable especially as we have to prioritise work this close to mainnet. 

Can you please update the issue description with:
What you mean by 'the environment is not scalable and predictable'
How this relates to util functions 
And what the plan is to resolve that

Thanks

Kishan-Dhakan:
@YarikRevich please update the issue description, thanks

YarikRevich:
@Kishan-Dhakan In API tests we have endpoint_test.go file. There are lots of functions used for API calls(V1GetBalance and so on). Those functions use zerochain.go file, where calls to specific services happen. Also, almost in each test file there are tx call functions, which are repeatable, and just make code unclear. In this issue I propose to create one package "client", where there will be both API calls and Tx calls without deviding on like "zerochain.go" and "endpoint_test.go" and spreading tx call logic across test files

stewartie4:
Spreading the logic across files is just good separation of concerns, we don't want to get into a situation where all logic lives in 1 giant procedural file
The intent is also one of abstraction so a function called "register wallet" may generate a mnemonic, call client put and verify that call. rather than having everything explicitly defined in each test

PR #416, opened 2022-09-20
Feature/refactor tests

YarikRevich:
Refers to issue #408


YarikRevich:
After refactoring it takes only 3 minutes to process all the API tests

internal/api/model/api.go
```diff
@@ -0,0 +1,479 @@
+package model
+
+import (
+	"encoding/json"
+	"github.com/0chain/gosdk/core/common"
```
stewartie4, line 5: we shouldn't be using gosdk for our models

internal/api/util/interaction/interaction.go
```diff
@@ -0,0 +1,26 @@
+package interaction
+
+import (
+	"crypto/rand"
+	"os"
+)
+
+// CreateFile creates a new file and fills it with random data
+func CreateFile(name string, size int64) (*os.File, error) {
```
stewartie4, line 9: this is unused

tests/api_tests/register_wallet_test.go
```diff
@@ -20,66 +16,19 @@ func TestRegisterWallet(t *testing.T) {
...
-	clientId := encryption.Hash(publicKeyBytes)
-
-	require.Nil(t, err, "Unexpected error [%s] occurred registering wallet with http response [%s]", err, httpResponse)
-	require.NotNil(t, registeredWallet, "Registered wallet was unexpectedly nil! with http response [%s]", httpResponse)
-	require.Equal(t, endpoint.HttpOkStatus, httpResponse.Status())
-	require.Equal(t, registeredWallet.ClientID, clientId)
-	require.Equal(t, registeredWallet.ClientKey, keyPair.PublicKey.SerializeToHexStr())
-	require.Greater(t, registeredWallet.MustConvertDateCreatedToInt(), 0, "Creation date is an invalid value!")
-	require.NotNil(t, registeredWallet.Version)
+		wallet, resp, err := apiClient.V1ClientPut(model.ClientPutRequest{}, client.HttpOkStatus)
```
stewartie4, line 19: You've removed a really important layer of abstraction - the caller should have a human readable function for readability eg. "registerWallet" rather than having to craft a V1ClientPut request every time

Same for most changes made, this leads to very hard to follow tests, such as https://github.com/0chain/system_test/pull/416/files#diff-c301310768a4fa7df5a7ee9b1f40831ac07fbdfba9c5443f1f019a1876c96850

internal/api/util/client/api.go
```diff
@@ -0,0 +1,571 @@
...
+
+	mnemonics := crypto.GenerateMnemonics()
+	keyPair := crypto.GenerateKeys(mnemonics)
+	publicKeyBytes, err := hex.DecodeString(keyPair.PublicKey.SerializeToHexStr())
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	if clientPutRequest.ClientID == "" {
+		clientPutRequest.ClientID = encryption.Hash(publicKeyBytes)
```
stewartie4, line 252: you're assuming if someone calls this function without a client id, that one should be generated.
This can lead to hard to debug tests,
It is not the responsibility of this function to do that, this function should be provided with all it needs

tests/api_tests/register_wallet_test.go
stewartie4, line 19: This test no longer has any visible test setup, and all of the low level assertions have been removed 
Tests should be explicit

internal/api/util/client/api.go
```diff
@@ -0,0 +1,571 @@
...
+	healthySharders := c.getHealthyShaders(networkDNSResponse.Sharders)
+	if len(healthySharders) == 0 {
+		return ErrNoShadersHealth
+	}
+
+	c.NetworkHealthResources.Sharders = healthySharders
+	return nil
+}
+
+func (c *APIClient) executeForServiceProvider(url string, executionRequest model.ExecutionRequest, method int) (*resty.Response, error) { //nolint
```
stewartie4, line 151: nice tidy up

internal/api/util/client/api.go
```diff
@@ -0,0 +1,571 @@
...
+		formattedURL := urlBuilder.MustShiftParse(serviceProvider).String()
+
+		newResp, err := c.executeForServiceProvider(formattedURL, executionRequest, method)
+		if err != nil {
+			errors = append(errors, err)
+			log.Println(err)
+			continue
+		}
+
+		if newResp.StatusCode() == executionRequest.RequiredStatusCode {
```
stewartie4, line 208: as far as I can see, you've removed the ability to categorise responses based on a function, this was really useful eg. when it came to consensus operations

stewartie4 (changes requested):
Partial review

stewartie4:
Useful logging has been removed:
Before:
image
After:
image

internal/api/util/client/api.go
```diff
@@ -0,0 +1,571 @@
...
+
+func (c *APIClient) V1TransactionPut(internalTransactionPutRequest model.InternalTransactionPutRequest, requiredStatusCode int) (*model.TransactionPutResponse, *resty.Response, error) { //nolint
+	var transactionPutResponse *model.TransactionPutResponse
+
+	data, err := json.Marshal(internalTransactionPutRequest.TransactionData)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	internalTransactionPutRequest.Wallet.IncNonce()
```
stewartie4, line 298: nonce should be incremented on a successful confirmation, not before a request is sent - if the request fails to send, the nonce value will be incorrect for the next request

internal/api/util/client/api.go
```diff
@@ -0,0 +1,571 @@
...
+		if err != nil {
+			log.Fatalln(err)
+		}
+
+		if resp.StatusCode() != requiredStatusCode {
+			return false
+		}
+
+		if transactionGetConfirmationResponse.Status != requiredTransactionStatus {
+			log.Println(transactionGetConfirmationResponse.Transaction.TransactionOutput)
```
stewartie4, line 381: we should be using testing.log instead of logger

internal/api/util/client/api.go
```diff
@@ -0,0 +1,571 @@
...
+		serviceProviders = c.NetworkHealthResources.Sharders
+	}
+
+	for _, serviceProvider := range serviceProviders {
+		formattedURL := urlBuilder.MustShiftParse(serviceProvider).String()
+
+		newResp, err := c.executeForServiceProvider(formattedURL, executionRequest, method)
+		if err != nil {
+			errors = append(errors, err)
+			log.Println(err)
```
stewartie4, line 204: this is already logged elsewhere

internal/api/util/client/api.go
```diff
@@ -0,0 +1,571 @@
...
+			continue
```
stewartie4, line 205: nice catch

internal/api/util/client/api.go
```diff
@@ -0,0 +1,571 @@
...
+		resp   *resty.Response
+		errors []error
+	)
+
+	var expectedExecutionResponseCounter, notExpectedExecutionResponseCounter int
+
+	var serviceProviders []string
+
+	switch serviceProviderType {
+	case MinerServiceProvider:
```
stewartie4, line 192: generally prefer polymorphism over if statements

stewartie4 (changes requested):
I like what you're trying to do, but we want to do it while keeping the flexibility and readability of the existing approach

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
```diff
@@ -4,8 +4,11 @@ import (
 	"bytes"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
+	"github.com/0chain/gosdk/core/encryption"
```
stewartie4, line 8: can we try not to use gosdk unless absolutely necessary

internal/api/util/crypto/crypto.go
```diff
@@ -55,50 +54,36 @@ func GenerateKeys(t *testing.T, mnemonic string) *model.KeyPair {
...
-	var hashData = blankIfNil(request.CreationDate) + ":" +
-		blankIfNil(request.TransactionNonce) + ":" +
-		blankIfNil(request.ClientId) + ":" +
-		blankIfNil(request.ToClientId) + ":" +
-		blankIfNil(request.TransactionValue) + ":" +
-		Sha3256([]byte(request.TransactionData))
-
-	request.Hash = Sha3256([]byte(hashData))
-}
+func HashOfFileSHA1(src *os.File) (string, error) {
```
stewartie4, line 67: the functions newconnectionid and HashOfFileSHA1 are unused

internal/api/util/crypto/crypto.go
```diff
@@ -55,50 +54,36 @@ func GenerateKeys(t *testing.T, mnemonic string) *model.KeyPair {
...
+	h := sha1.New()
+	if _, err := io.Copy(h, src); err != nil {
+		return "", err
+	}
+	if _, err := src.Seek(0, io.SeekStart); err != nil {
+		return "", err
+	}
 
-func Sha3256(publicKeyBytes []byte) string {
```
stewartie4, line 85: why are these functions being removed? separation of concerns means we should have a utility to hash a given object, not have the object hash itself

stewartie4:
> After refactoring it takes only 3 minutes to process all the API tests

Nice! how did you manage that?

I did notice some tests are taking roughly double the time now though:
Before:
image


After
image

Maybe error scenarios are being retried before passing?

internal/api/util/interaction/interaction.go
YarikRevich: It's used for my blobber token accounting test

internal/api/util/crypto/crypto.go
YarikRevich: User Acceptance tests should represent possible usage of components from the customer point of view. Customer won't generate this hash himself, it's absurd. There should be common package with all functions needed for encryption

internal/api/util/crypto/crypto.go
YarikRevich: They are used for blobber token accounting tests

YarikRevich: I'll add a PR for them, after this PR merged

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
YarikRevich: This encryption package should be moved to common package

internal/api/util/client/api.go
YarikRevich: This function after request processing calls a concensus functions to select that error which is the most frequent in "errors" slice

internal/api/util/client/api.go
YarikRevich: Yeah, should remove that log.Println

internal/api/util/client/api.go
YarikRevich: If we use testing.log, we should pass testing.T pointer everywhere, it doesn't only look bad, but also may allocate more memory

tests/api_tests/register_wallet_test.go
YarikRevich: `wallet, resp, err := apiClient.V1ClientPut(model.ClientPutRequest{}, client.HttpOkStatus)
		require.Nil(t, err)
		require.NotNil(t, resp)
		require.NotNil(t, wallet)

		require.NotZero(t, len(wallet.Keys))

		publicKeyBytes, err := hex.DecodeString(wallet.RawKeys.PublicKey.SerializeToHexStr())
		require.Nil(t, err)

		require.Equal(t, wallet.ClientID, encryption.Hash(publicKeyBytes))
		require.Equal(t, wallet.ClientKey, wallet.MustGetKeyPair().PublicKey)
		require.NotZero(t, wallet.MustConvertDateCreatedToInt(), "creation date is an invalid value!")
		require.NotZero(t, wallet.Version)`



We have

YarikRevich:
@stewartie4 Well, I propose a compromise. I will create "wrappers" for explicit API requests, but it won't be spared across different files in "api_tests" package. It will be in the same "client" package, where all explicit API requests defined.

tests/api_tests/register_wallet_test.go
YarikRevich: Well, I propose a compromise. I will create "wrappers" for explicit API requests, but it won't be spared across different files in "api_tests" package. It will be in the same "client" package, where all explicit API requests defined.

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
Kishan-Dhakan: I agree, let's not use gosdk here please. If signature or key generation is broken in gosdk then it will cause failures here as well, we don't want that.

tests/cli_tests/zboxcli_collect_rewards_test.go
```diff
@@ -60,7 +60,7 @@ func TestCollectRewards(t *testing.T) {
 		// Use all 6 blobbers
 		allocationID := setupAllocationAndReadLock(t, configPath, map[string]interface{}{
 			"size":   allocSize,
-			"tokens": 1,
+			"tokens": 0.2,
```
Kishan-Dhakan, line 63: Revert, also might need to investigate why rewards are not showing for 0.2 tokens. Maybe create a separate ticket or raise it in slack

Kishan-Dhakan (changes requested):

internal/api/util/client/api.go
stewartie4: Neither points matter when it comes to test readability 
testing.log is the standard, let's use it please

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
stewartie4: @YarikRevich regardless, we should be writing these tests like an end user, with no 0chain helper code to ensure nothing breaks as Kishan mentions

internal/api/util/crypto/crypto.go
stewartie4: The user of the 0chain API will indeed generate the hash themselves
Anyway, this is a question of internal code structure, not how the end user will call the API

internal/api/util/crypto/crypto.go
stewartie4: let's remove any unused code and add it on the PR it is required for

internal/api/util/interaction/interaction.go
stewartie4: I don't see that in this PR?

tests/cli_tests/zboxcli_collect_rewards_test.go
YarikRevich: Sure, will return 1

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
YarikRevich: Ok

tests/api_tests/register_wallet_test.go
stewartie4: No visible test setup:
The first line calls v1ClientPut with no test data, that is all setup internally to that function

low level assertions have been removed:
require.Nil(t, err, "Unexpected error [%s] occurred registering wallet with http response [%s]", err, rawHttpResponse)
require.NotNil(t, registeredWallet, "Registered wallet was unexpectedly nil! with http response [%s]", rawHttpResponse)
require.Equal(t, endpoint.HttpOkStatus, rawHttpResponse.Status())

Not just this test, many tests have been modified in this way

stewartie4:
> @stewartie4 Well, I propose a compromise. I will create "wrappers" for explicit API requests, but it won't be spared across different files in "api_tests" package. It will be in the same "client" package, where all explicit API requests defined.

Sounds good

tests/api_tests/register_wallet_test.go
stewartie4: sounds good

internal/api/model/sdk.go
```diff
@@ -0,0 +1,27 @@
+package model
+
+import (
+	"github.com/0chain/gosdk/core/transaction"
```
stewartie4, line 4: Again, don't use gosdk

internal/api/util/client/api.go
```diff
@@ -0,0 +1,696 @@
...
+		},
+		HttpOkStatus)
+	require.Nil(t, err)
+	require.NotNil(t, resp)
+
+	return scRestGetAllocationBlobbersResponse
+}
+
+// GetAllocationWrapper does not provide deep test of used components
+func (c *APIClient) GetAllocationWrapper(t *testing.T, allocationID string) *model.SCRestGetAllocationResponse {
```
stewartie4, line 668: I'd like these abstraction to be as it was in the original framework:
https://github.com/0chain/system_test/blob/af839eb97243b777ea589e691c28cf4eccdf9f1e/tests/api_tests/create_allocation_test.go#L66
Where it returns the serialised object, the raw http response and any error that was encountered 
There should then be another function that performs basic assertions on the first function for convenience eg.
https://github.com/0chain/system_test/blob/af839eb97243b777ea589e691c28cf4eccdf9f1e/tests/api_tests/create_allocation_test.go#L56

internal/api/util/client/api.go
```diff
@@ -0,0 +1,696 @@
...
+func (c *APIClient) CreateStakePoolWrapper(t *testing.T) {
+	t.Log("Create stake pool...")
+}
+
+// UpdateAllocationWrapper does not provide deep test of used components
+func (c *APIClient) UpdateAllocationWrapper(t *testing.T) {
+	t.Log("Update allocation...")
+}
+
+func (c *APIClient) GetAllocationBlobbersWrapper(t *testing.T, wallet *model.Wallet) model.SCRestGetAllocationBlobbersResponse {
```
stewartie4, line 652: No need for "Wrapper" at the end - every function is a later of abstraction and a "wrapper" of sorts

internal/api/util/client/sdk.go
```diff
@@ -0,0 +1,41 @@
+package client
+
+import (
+	"github.com/0chain/gosdk/core/conf"
```
stewartie4, line 4: remove gosdk

internal/api/util/client/sdk.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/0chain/gosdk/zboxcore/sdk"
```
stewartie4, line 5: remove gosdk

internal/api/util/wait/wait.go
```diff
@@ -6,12 +6,12 @@ import (
 
 // PoolImmediately pools passed function for a certain amount of time
 func PoolImmediately(duration time.Duration, callback func() bool) {
-	ticker := time.NewTicker(time.Second * 3)
+	ticker := time.NewTicker(time.Millisecond * 500)
```
stewartie4, line 9: Add logging to this polling

tests/api_tests/add_blobber_test.go
```diff
@@ -58,147 +63,141 @@ func TestAddBlobber(t *testing.T) {
...
 
-		allocationUpdate := getAllocationUpdate(allocation.ID, strconv.Itoa(rand.Intn(10)), "")
-		updateAllocationTransactionResponse, confirmation := updateAllocation(t, registeredWallet, keyPair, allocationUpdate)
-		require.NotNil(t, updateAllocationTransactionResponse)
-		require.Equal(t, endpoint.TxUnsuccessfulStatus, confirmation.Status)
-
-		allocation = getAllocation(t, createAllocationTransactionResponse.Entity.Hash)
-		require.NotNil(t, allocation)
-
+		updateAllocationTransactionPutResponse, resp, err := apiClient.V1TransactionPut(
```
stewartie4, line 119: updateAllocation abstraction required, and anywhere else we have this verbose call to put a transaction

tests/api_tests/flaky___broken_scenarios_test.go
```diff
@@ -23,63 +22,76 @@ func Test___BrokenScenariosRegisterWallet(t *testing.T) {
...
-
-		walletResponse, httpResponse, err := v1ClientPut(t, walletRequest, endpoint.ConsensusByHttpStatus("400 Bad Request"))
-
-		require.Nil(t, walletResponse, "Expected returned wallet to be nil but was [%s] with http response [%s]", walletResponse, httpResponse)
-		require.NotNil(t, err, "Expected error when registering wallet but was nil.")
-		require.Equal(t, "400 Bad Request", httpResponse.Status())
+		mnemonics := crypto.GenerateMnemonics()
+		expectedKeyPair := crypto.GenerateKeys(mnemonics)
+
+		wallet, resp, err := apiClient.V1ClientPut(
```
stewartie4, line 55: again, abstraction needed

internal/api/util/client/api.go
```diff
@@ -0,0 +1,696 @@
+package client
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"github.com/0chain/gosdk/core/sys"
```
stewartie4, line 7: remove gosdk

stewartie4 (changes requested):
Some previous comments still to be resolved, still some go sdk usage, still some abstraction needed

tests/api_tests/add_blobber_test.go
YarikRevich: We shouldn't use abstractions here, because we test "update allocation" endpoint functionality. I wrote in the comment of each "wrapper", that it doesn't provide a deep test for the calls it wraps

tests/api_tests/flaky___broken_scenarios_test.go
YarikRevich: In this test we test "client put" endpoint functionality, so we shouldn't use here abstractions, because "wrappers" don't provide deep test, I wrote that

internal/api/util/client/api.go
YarikRevich: No, no, no, It's a total antipattern, to provide deep testing, we need, to write a raw API call, "wrappers" are used for those tests, where the functionality they wrap shouldn't be tested

internal/api/util/wait/wait.go
YarikRevich: Logging is provided in the callback this pool calls

internal/api/util/client/sdk.go
YarikRevich: This is sdk model. It's used for sdk client, which is used util we have upload API

internal/api/util/client/api.go
YarikRevich: We need "wrapper" prefix, because all the "wrappers" are located in "client" package. If they were located in "wrapper" package, well, there wouldn't be any reasons to use that prefix. But the key is that we implemented this "wrappers" not in another package, but even in the same struct, so we should difference them

internal/api/util/client/sdk.go
YarikRevich: This is sdk model. It's used for sdk client, which is used util we have upload API

tests/api_tests/register_wallet_test.go
YarikRevich: We don't these kind of assertions with these "messages", in the function, where API call is provided, we can see all the important logs

internal/api/util/interaction/interaction.go
YarikRevich: It's another PR, which depends on this one

internal/api/util/client/api.go
YarikRevich: Fixed

internal/api/model/sdk.go
YarikRevich: It's used for sdk client

internal/api/util/client/sdk.go
stewartie4: We don't use gosdk in our API tests, the reason is irrelevant really 
 It's a principle of test design, this needs removed

internal/api/util/wait/wait.go
stewartie4: you can't know that, the callback could be anything 
Surely we need polling for
- call success
- call failed 
- call timed out

tests/api_tests/register_wallet_test.go
stewartie4: function call still needed here

tests/api_tests/flaky___broken_scenarios_test.go
stewartie4: What do you mean by "deep test"?
Previously we had 2 functions, one performing no assertions and the other performing basic assertions incase it is used in test setup elsewhere

This is already decided on master for a reason, let's stick to that

tests/api_tests/add_blobber_test.go
stewartie4: same as other comment

internal/api/util/client/sdk.go
stewartie4: We don't use gosdk in our API tests, the reason is irrelevant really
It's a principle of test design, this needs removed

internal/api/util/client/api.go
stewartie4: They're not wrappers, they're just function calls - layers of abstraction.
is os.LoadFiile os.LoadFileWrapper? no because we understand how abstraction works 
"Wrapper" adds nothing to what the function does in terms of readability

internal/api/util/client/api.go
stewartie4: Tests need to be readable at a glance 

`wallet := apiClient.RegisterWalletWrapper(t)
		apiClient.ExecuteFaucetWrapper(t, wallet)

		scRestGetAllocationBlobbersResponse, resp, err := apiClient.V1SCRestGetAllocationBlobbers(
			&model.SCRestGetAllocationBlobbersRequest{
				ClientID:  wallet.ClientID,
				ClientKey: wallet.ClientKey,
			},
			client.HttpOkStatus)
		require.Nil(t, err)
		require.NotNil(t, resp)
		require.NotNil(t, scRestGetAllocationBlobbersResponse)

		createAllocationTransactionPutResponse, resp, err := apiClient.V1TransactionPut(
			model.InternalTransactionPutRequest{
				Wallet:          wallet,
				ToClientID:      client.StorageSmartContractAddress,
				TransactionData: model.NewCreateAllocationTransactionData(scRestGetAllocationBlobbersResponse),
				Value:           tokenomics.IntToZCN(0.1),
			},
			client.HttpOkStatus)
		require.Nil(t, err)
		require.NotNil(t, resp)
		require.NotNil(t, createAllocationTransactionPutResponse)

		createAllocationTransactionGetConfirmationResponse, resp, err := apiClient.V1TransactionGetConfirmation(
			model.TransactionGetConfirmationRequest{
				Hash: createAllocationTransactionPutResponse.Entity.Hash,
			},
			client.HttpOkStatus,
			client.TxSuccessfulStatus)
		require.Nil(t, err)
		require.NotNil(t, resp)
		require.NotNil(t, createAllocationTransactionGetConfirmationResponse)

		scRestGetAllocation, resp, err := apiClient.V1SCRestGetAllocation(
			model.SCRestGetAllocationRequest{
				AllocationID: createAllocationTransactionPutResponse.Entity.Hash,
			},
			client.HttpOkStatus)
		require.Nil(t, err)
		require.NotNil(t, resp)
		require.NotNil(t, scRestGetAllocation)
	})`

The above is the diametric opposite of that.
I don't want to argue on every point Yarik, i'm only asking for a bit of abstraction so our tests are readable and maintainable

internal/api/model/sdk.go
stewartie4: remove it from the sdk client then, we shouldn't have it in our tests

internal/api/util/crypto/crypto.go
YarikRevich: Ok

internal/api/util/client/api.go
YarikRevich: "os" package is a part of Golang, but the Golang language itself is the whole abstraction to OS low level C API. That's why this Golang function has no "wrapper" suffix. If such a tiny function was defined in low level C API, then it should have some prefix, which will define it as a high level function for a low level logic

internal/api/util/wait/wait.go
YarikRevich: Ok, I'll provide it

internal/api/util/client/api.go
stewartie4: So you're saying most high level functions should have the suffix "wrapper"?
That's just not the case, search any codebase and you'll not find a single instance of that 
I've never heard such a thing in all my years - unless i'm misunderstanding you it seems completely superfluous

internal/api/util/client/api.go
Kishan-Dhakan: @YarikRevich is this done selectMostFrequentError now? If not please re-instate the code that categorised responses

internal/api/util/client/api.go
Kishan-Dhakan: @YarikRevich I think what you are trying to do is, if someone calls `registerWalletWrapper` they do not have to do the entire test that would have been something like `t.Run("Register wallet via API should work")`, instead just do some basic assertions. I assume this is what you mean by deep test.

Deeptesting every step in all tests can provide fast failures but also take longer. I think it's a good idea, but `Wrapper` does not do anything to explain this. Please rename these from `RegisterWalletWrapper` to something like `RegisterWalletWithoutAssertions` or something. Wrapper does not explain anything

internal/api/util/client/api.go
Kishan-Dhakan: Same comment as above, rename from Wrapper to something else.

internal/api/util/crypto/crypto.go
Kishan-Dhakan: Please remove this and merge later in respective PR

internal/api/util/wait/wait.go
Kishan-Dhakan: Has this been added?

tests/api_tests/add_blobber_test.go
Kishan-Dhakan: We need an abstraction here but not the "Wrapper" one. Simply make an abstraction like `updateAllocation`, coupled with `updateAllocationWithoutAssertions` it should make sense.

tests/api_tests/flaky___broken_scenarios_test.go
Kishan-Dhakan: @YarikRevich , again, you don't need to use your `Wrapper` (which I call `withoutAssertions`) here, you can also provide abstraction using a new method :)

tests/api_tests/register_wallet_test.go
Kishan-Dhakan: Please put the abstraction back in @YarikRevich (not the wrapper, the deep test one)

Kishan-Dhakan: Same as above, put a deep test abstraction here

Kishan-Dhakan (changes requested):

internal/api/util/crypto/crypto.go
Kishan-Dhakan: @YarikRevich

internal/api/util/crypto/crypto.go
Kishan-Dhakan: @YarikRevich

Kishan-Dhakan (changes requested):

internal/api/util/crypto/crypto.go
YarikRevich: @Kishan-Dhakan I've added sha3256 function?

Kishan-Dhakan (approved):

tests/cli_tests/zwalletcli_send_and_balance_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
-	apimodel "github.com/0chain/system_test/internal/api/model"
+	//apimodel "github.com/0chain/system_test/internal/api/model"
```
stewartie4, line 15: commented out code

internal/api/util/wait/wait.go
```diff
@@ -20,11 +21,15 @@ primary:
 			break primary
 		default:
 			if callback() {
+				log.Println("Wait pool callback has succeed")
```
stewartie4, line 24: should use t.log

internal/api/model/api.go
```diff
@@ -0,0 +1,483 @@
...
+	PrivateKey bls.SecretKey
+}
+
+func (w *Wallet) IncNonce() {
+	w.Nonce++
+}
+
+func (w *Wallet) MustGetKeyPair() KeyPair {
+	if len(w.Keys) == 0 {
+		log.Fatalln("wallet has no keys")
```
stewartie4, line 76: should use t.log

internal/api/util/client/api.go
```diff
@@ -0,0 +1,865 @@
...
+		resp, err = c.httpClient.R().SetHeaders(executionRequest.Headers).SetFormData(executionRequest.FormData).SetBody(executionRequest.Body).Post(url)
+	case HttpGETMethod:
+		resp, err = c.httpClient.R().SetHeaders(executionRequest.Headers).SetQueryParams(executionRequest.QueryParams).Get(url)
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", url, ErrGetFromResource)
+	}
+
+	log.Printf("%s returned %s with status %s", url, resp.String(), resp.Status())
```
stewartie4, line 173: should use t.log

internal/api/util/client/api.go
```diff
@@ -0,0 +1,865 @@
...
+func (c *APIClient) V1ClientPut(clientPutRequest model.ClientPutRequest, requiredStatusCode int) (*model.Wallet, *resty.Response, error) { //nolint
+	var clientPutResponse *model.ClientPutResponse
+
+	urlBuilder := NewURLBuilder().SetPath(ClientPut)
+
+	mnemonics := crypto.GenerateMnemonics()
+	keyPair := crypto.GenerateKeys(mnemonics)
+	publicKeyBytes, err := hex.DecodeString(keyPair.PublicKey.SerializeToHexStr())
+	if err != nil {
+		log.Fatalln(err)
```
stewartie4, line 250: should use t.log

internal/api/util/client/api.go
```diff
@@ -0,0 +1,865 @@
...
+
+	return wallet, resp, err
+}
+
+func (c *APIClient) V1TransactionPut(internalTransactionPutRequest model.InternalTransactionPutRequest, requiredStatusCode int) (*model.TransactionPutResponse, *resty.Response, error) { //nolint
+	var transactionPutResponse *model.TransactionPutResponse
+
+	data, err := json.Marshal(internalTransactionPutRequest.TransactionData)
+	if err != nil {
+		log.Fatalln(err)
```
stewartie4, line 299: t.log

internal/api/util/client/api.go
```diff
@@ -0,0 +1,865 @@
...
+		transactionPutRequest.CreationDate,
+		transactionPutRequest.TransactionNonce,
+		transactionPutRequest.ClientId,
+		transactionPutRequest.ToClientId,
+		transactionPutRequest.TransactionValue,
+		crypto.Sha3256([]byte(transactionPutRequest.TransactionData)))))
+
+	hashToSign, err := hex.DecodeString(transactionPutRequest.Hash)
+	if err != nil {
+		log.Fatalln(err)
```
stewartie4, line 330: t.log

tests/cli_tests/zboxcli_create_allocation_free_storage_test.go
```diff
@@ -358,14 +363,29 @@ func freeAllocationAssignerTxn(t *testing.T, from, assigner *climodel.WalletFile
...
+		txn.CreationDate,
+		txn.TransactionNonce,
+		txn.ClientId,
+		txn.ToClientId,
+		txn.TransactionValue,
+		crypto.Sha3256([]byte(txn.TransactionData)))))
+
+	hashToSign, err := hex.DecodeString(txn.Hash)
+	if err != nil {
+		log.Fatalln(err)
```
stewartie4, line 382: here and 14 other places

stewartie4 (changes requested):
looks great, some small comments remaining

stewartie4:
> Useful logging has been removed: Before: image After: image

this is still an issue

stewartie4:
> > After refactoring it takes only 3 minutes to process all the API tests
> 
> Nice! how did you manage that?
> 
> I did notice some error scenarios are taking roughly double the time now though: Before: image
> 
> After image image
> 
> Maybe error scenarios are being retried before passing?

did you look into this?

internal/api/model/api.go
YarikRevich: Don't you think that it'll be good to separate logging of tests and util functions?

YarikRevich:
@stewartie4 Reduced the usage of 'log' package and replaced it, as you said, to 't.Log', but left some log.Print in the functions which are called from "MainTest" and for 'url' calls. What do you think about it?

YarikRevich:
Fixed performance issue, check it out

stewartie4 (approved):

PR #438, opened 2022-10-06
livestream download test

SinanEkm:


tests/cli_tests/zboxcli_download_livestream_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+		localfolderForDownload := filepath.Join(os.TempDir()+"download", escapedTestName(t))
+		localpathForDownload := filepath.Join(localfolderForDownload, "down.m3u8")
+		err = os.MkdirAll(localpathForDownload, os.ModePerm)
+		require.Nil(t, err, "Error in creating the folders", localpathForDownload)
+
+		defer os.RemoveAll(localfolder)
+		defer os.RemoveAll(localfolderForDownload)
+
+		wg := new(sync.WaitGroup)
```
Kishan-Dhakan, line 59: Minor: Can use wg := &sync.WaitGroup{}, it's cleaner than using new()

tests/cli_tests/zboxcli_download_livestream_test.go
```diff
@@ -0,0 +1,163 @@
...
+		t.Skipf("Youtube live streams are not available right now!")
+	}
+
+	t.Run("Downloading youtube feed to allocation should work", func(t *testing.T) {
+		t.Parallel()
+
+		output, err := registerWallet(t, configPath)
+		require.Nil(t, err, "failed to register wallet", strings.Join(output, "\n"))
+
+		output, err = executeFaucetWithTokens(t, configPath, 2.0)
```
Kishan-Dhakan, line 35: Can you use  `_ = initialiseTest(t)` here instead?

tests/cli_tests/zboxcli_download_livestream_test.go
SinanEkm: Yeah, sure.

tests/cli_tests/zboxcli_download_livestream_test.go
SinanEkm: In which line exactly? Like, after the t.Parallel()?

Kishan-Dhakan: Replace the register and faucet steps into one abstraction. Check the function definition of initialiseTest(t) and replace appropriately

SinanEkm:
I have planned to test three cases such as;

1. Downloading the Youtube feed to allocation
2. Downloading local webcam feed
3. Downloading the feed to allocation with a delay flag

At the end of the day, 1. and 3. tests are working fine. However, I couldn't figure out how to test the 2. test case. It needs to open the webcam and stream it to the allocation but, it couldn't open my webcam and stream it to the allocation. Although if it could open my webcam, I am not sure how it can be done in GitHub actions.

On the other hand, I also added a few controls such as if uploaded files and downloaded files are the same (by using md5 hashing), and if the download folder has any .ts and .m3u8 files.

SinanEkm:
- Added the webcam test, it works fine so far
- Changed the wait times since the existing times (45 and 60 seconds) were not enough for the tasks and caused to fail tests.

SinanEkm:
Weirdly tests are sometimes passing and sometimes don’t. I tracked this issue on my local and have been tracking the Github actions at the same time. I am not sure but, I guess running tasks in parallel and doing upload/download concurrently causes this unstably. So, I did a few tests on my local with different settings and, running tests sequentially solves the problem but it is not effective. Maybe we can run two of them in parallel and one sequentially. Not sure at this point and waiting for the test results of the Github actions.

tests/cli_tests/zboxcli_download_livestream_test.go
```diff
@@ -0,0 +1,545 @@
...
+		}
+
+		walletOwner := escapedTestName(t) + "_wallet"
+
+		_ = initialiseTest(t, walletOwner, true)
+
+		output, err := createNewAllocation(t, configPath, createParams(map[string]interface{}{
+			"lock": 0.5,
+		}))
+		t.Log(output)
```
Kishan-Dhakan, line 45: Can we remove this log?

Kishan-Dhakan (approved):

tests/cli_tests/zboxcli_download_livestream_test.go
SinanEkm: Ok, I removed it.

Kishan-Dhakan (approved):

Issue #338, opened 2022-07-23
Blobber: /v1/file/referencepath & objecttree API tests and doc review


stewartie4:
Investigate and Implement system acceptance tests directly against the blobber  /v1/file/objectpath  /v1/file/referencepath and objecttree APIs
Example test cases for each endpoint are below but **this is not a complete list.**
Also create your own, thinking about security, param validation, error messaging, performance etc.

valid request(s)
no allocation id
invalid allocation id
someone else's allocation
invalid owner
no paths
null paths
invalid path
expired allocation
invalid signature 
empty body/params/form
  
Following the[ test style guide](https://docs.google.com/document/d/12RtuT89FG7PycsOno7312yQoJ4nzzfs0xLBEDA38Q5k/edit?usp=sharing) and the style of the existing API tests

Also check that this endpoint is list on api.0chain.net and is up to date with a recent example

Kishan-Dhakan:
/GetObjectPath API is removed as from blobbers in https://github.com/0chain/blobber/pull/863.

PR #454, opened 2022-10-14
Fix API tests flakiness

Kishan-Dhakan:
- Closes https://github.com/0chain/system_test/issues/338


tests/api_tests/get_blobberFileRef_test.go
```diff
@@ -11,20 +11,11 @@ import (
 )
 
 func TestBlobberFileRefs(t *testing.T) {
-	t.Parallel()
```
nlok5923, line 14: Can't we still make this test run parallely along with making gosdk using one wallet at a time.

tests/api_tests/main_test.go
```diff
@@ -26,5 +30,11 @@ func TestMain(m *testing.M) {
 	sdkClient = client.NewSDKClient(parsedConfig.BlockWorker)
 	apiClient = client.NewAPIClient(parsedConfig.BlockWorker)
 
+	t := new(testing.T)
+
+	sdkWalletMnemonics = crypto.GenerateMnemonics(t)
+	sdkWallet = apiClient.RegisterWalletForMnemonic(t, sdkWalletMnemonics)
+	sdkClient.SetWallet(t, sdkWallet, sdkWalletMnemonics)
+
```
YarikRevich, line 38: I think this decesion to register a wallet only once for sdk client is a good way to go, but also it's better to make this "initialization" in a separate function. It's a minor change

tests/api_tests/get_blobber_rewards_test.go
```diff
@@ -5,33 +5,23 @@ import (
 	"time"
 
 	"github.com/0chain/system_test/internal/api/util/client"
-	"github.com/0chain/system_test/internal/api/util/crypto"
 	"github.com/0chain/system_test/internal/api/util/wait"
 	"github.com/stretchr/testify/require"
 )
 
 func TestBlobberRewards(t *testing.T) {
-	t.Parallel()
```
YarikRevich, line 14: You can start using parallel execution of the tests, because you synchronized all actions in sdkClient.

YarikRevich (changes requested):

internal/api/util/client/sdk.go
```diff
@@ -75,10 +76,7 @@ func (c *SDKClient) UploadFile(t *testing.T, allocationID string) string {
 	}
 
 	defer func(name string) {
-		err := os.RemoveAll(name)
-		if err != nil {
-
-		}
+		_ = os.RemoveAll(name)
```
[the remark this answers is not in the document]
Kishan-Dhakan: This wasn't being checked at all till now. It was just an empty if block. This is more of a lint fix than logic change.

tests/api_tests/main_test.go
Kishan-Dhakan: Okay, agreed. Will refactor

tests/api_tests/get_blobberFileRef_test.go
YarikRevich: I already left a comment on this. Yes, we can, cause @Kishan-Dhakan has fixed synchronization of actions for sdk client.

YarikRevich (approved):

nlok5923 (approved):

internal/api/util/client/sdk.go
```diff
@@ -40,7 +40,7 @@ func NewSDKClient(blockWorker string) *SDKClient {
 
 func (c *SDKClient) SetWallet(t *testing.T, wallet *model.Wallet, mnemonics string) {
 	c.Mutex.Lock()
-
+	defer c.Mutex.Unlock()
```
stewartie4, line 43: I hit these same issues and was investigating before I left.
Everything is 'static' (from a traditional language point of view) and global - we can use locks to kind of but not fully avoid the issues, but the underlying problem is still there.
Why can we not have multiple instances of this sdk client? 

For example even with locks, setWallet can still be called in one thread and upload can called in another, so you end up uploading using the incorrect wallet
Both calls will successfully establish a lock, but we still have state bleed between tests

stewartie4 (changes requested):
Look into multiple instantiation of the gosdk - if we can't do this then it is a serious deficiency in the gosdk itself - how do we ever expect anyone other than a hobbyist to use it?

tests/api_tests/main_test.go
stewartie4: every test should have it's own independent setup eg. wallet and allocation, else this will cause issues with token accounting tests

tests/api_tests/get_blobberFileRef_test.go
Kishan-Dhakan: It causes nonce related errors in running them in parallel. So I'm keeping them sequential for now. Since we plan on removing gosdk dependency soon anyhow.

Kishan-Dhakan:
> Look into multiple instantiation of the gosdk - if we can't do this then it is a serious deficiency in the gosdk itself - how do we ever expect anyone other than a hobbyist to use it?
@stewartie4 Created an issue for it here: https://github.com/0chain/gosdk/issues/642, since we can only test code as it is for now, we should get these tests in before mainnet.

stewartie4:
Agreed

stewartie4 (approved):

PR #460, opened 2022-10-22
WIP - General tidy up and CI improvements 

stewartie4:


internal/api/util/client/api.go
```diff
@@ -853,39 +853,15 @@ func (c *APIClient) UpdateAllocationBlobbers(t *testing.T, wallet *model.Wallet,
 	wallet.IncNonce()
 }
 
-func (c *APIClient) GetAllocationBlobbers(t *testing.T, wallet *model.Wallet, customBlobberRequirements *model.BlobberRequirements, requiredStatusCode int) *model.SCRestGetAllocationBlobbersResponse {
+func (c *APIClient) GetAllocationBlobbers(t *testing.T, wallet *model.Wallet, blobberRequirements model.BlobberRequirements, requiredStatusCode int) *model.SCRestGetAllocationBlobbersResponse {
 	t.Log("Get allocation blobbers...")
 
-	var blobberRequirements model.BlobberRequirements
-
-	if customBlobberRequirements != nil {
```
stewartie4, line 861: We should be explicit with our functions - it is not the function's responsibility to default to valid test data if none is supplied - this could cause a test bug or confusion

tests/cli_tests/zwalletcli_zcnbridge_ethereum-account-register_test.go
```diff
@@ -111,7 +112,7 @@ func getConfigDir() string {
 	var configDir string
 	curr, err := os.Getwd()
 	if err != nil {
-		panic(err)
+		log.Fatalln(err)
```
stewartie4, line 115: Mixed concerns is something that the go sdk seems to like!
The logger call here will also call os.exit(1) after logging which allows the test runner to record test results rather than a panic

internal/api/model/helper.go
```diff
@@ -0,0 +1,22 @@
+package model
+
```
stewartie4, line 2: Ideas re: the name of this file would be great 😆

YarikRevich (approved):
LGTM

nlok5923 (approved):

PR #462, opened 2022-10-27
443 broken tests on staging

SinanEkm:
All test cases are passing in my local so far. I also edited "Staking tokens against sharder should return intrests to wallet" case but, to be honest, I am not sure if it is the correct way of doing it or not so, I left the t.Skip() for now.


tests/cli_tests/zwalletcli_sharder_stake_test.go
```diff
@@ -150,28 +149,58 @@ func TestSharderStake(t *testing.T) {
...
+
+		// Stake tokens against this blobber
+		output, err = stakeTokens(t, configPath, createParams(map[string]interface{}{
+			"blobber_id": blobber.Id,
+			"tokens":     0.5,
+		}), true)
+		require.Nil(t, err, "Error staking tokens", strings.Join(output, "\n"))
+		require.Len(t, output, 1)
+		require.Regexp(t, regexp.MustCompile("tokens locked, txn hash: ([a-f0-9]{64})"), output[0])
+
```
Kishan-Dhakan, line 182: Why is there a need to stake tokens on a blobber for this test? The test name says sharder stake

tests/cli_tests/zwalletcli_mn_pool_info_test.go
```diff
@@ -13,6 +14,38 @@ import (
...
+		if sharder.ID != sharderNodeDelegateWallet.ClientID {
+			break
+		}
+	}
+
+	if _, err := os.Stat("./config/" + miner01NodeDelegateWalletName + "_wallet.json"); err != nil {
+		t.Skipf("miner node owner wallet located at %s is missing", "./config/"+miner01NodeDelegateWalletName+"_wallet.json")
+	}
+
+	miners := getMinersListForWallet(t, miner01NodeDelegateWalletName)
```
Kishan-Dhakan, line 37: Can you try using miner03NodeDelegateWalletName here instead of 01 and 02, seems they hit the max limit (20) in other tests

Kishan-Dhakan (changes requested):

stewartie4 (commented):
LGTM pending Kishan's comments

stewartie4 (approved):

tests/cli_tests/zwalletcli_sharder_stake_test.go
SinanEkm: Tbh, I have no idea. I just looked at similar test cases and tried something. But, as it turns out I was wrong. Deleted the lines I added and now it is passing the test.

tests/cli_tests/zwalletcli_mn_pool_info_test.go
SinanEkm: Ok, I changed it to `miner03NodeDelegateWalletName`. But still, it is not passing all the tests every time. It sometimes passes and sometimes not passing.

Kishan-Dhakan (approved):
LGTM

PR #486, opened 2022-11-29
/v1/block/get/latest_finalized_magic_block

SinanEkm:
3 test cases are added.


tests/api_tests/get_latest_finalized_magic_block_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"net/http"
+	"testing"
+)
+
+func TestGetLatestFinalizedMagicBlock(t *testing.T) {
+	t.Parallel()
+
+	hash, err := getCurrentHash(t)
+	require.Nil(t, err)
+	fmt.Println(hash)
```
Kishan-Dhakan, line 19: Remove log line

tests/api_tests/get_latest_finalized_magic_block_test.go
```diff
@@ -0,0 +1,108 @@
...
+		lfmb.Hash = ""
+		aa := model.LatestFinalizedMagicBlock{
+			LFMB: lfmb,
+		}
+		_, resp, err := apiClient.V1BlockGetLatestFinalizedMagicBlock(t, &aa, http.StatusOK)
+		require.Equal(t, resp.RawResponse.StatusCode, http.StatusOK)
+		require.Nil(t, err)
+	*/
+
+	resp, err := http.Get("https://test.0chain.net/sharder01/v1/block/get/latest_finalized_magic_block")
```
Kishan-Dhakan, line 89: Please replace this static URL and get it from Network service providers object

SinanEkm: Yes, I left this just temporarily for the PR.

internal/api/model/api.go
```diff
@@ -643,6 +644,10 @@ type SCRestGetStakePoolStatResponse struct {
 	Settings     StakePoolSettings           `json:"settings"`
 }
 
+type LatestFinalizedMagicBlock struct {
+	LFMB *block.Block
```
stewartie4, line 648: This uses a model from gosdk - we should not do this as if gosdk changes the tests will still pass

tests/api_tests/get_latest_finalized_magic_block_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/0chain/system_test/internal/api/model"
+	"github.com/stretchr/testify/require"
+	"net/http"
+	"testing"
+)
+
+func TestGetLatestFinalizedMagicBlock(t *testing.T) {
+	t.Parallel()
+
+	hash, err := getCurrentHash(t)
```
stewartie4, line 17: These lines should be part of test setup

tests/api_tests/get_latest_finalized_magic_block_test.go
```diff
@@ -0,0 +1,108 @@
...
+	require.Nil(t, err)
+	fmt.Println(hash)
+
+	t.Run("Not modified, set node lfmb", func(t *testing.T) {
```
stewartie4, line 21: Can this test case description be clearer?
Given x, when y, then z
for example

tests/api_tests/get_latest_finalized_magic_block_test.go
```diff
@@ -0,0 +1,108 @@
...
+		t.Parallel()
+		lfmb := &block.Block{}
+		lfmb.Hash = common.Key(hash)
+
+		aa := model.LatestFinalizedMagicBlock{
+			LFMB: lfmb,
+		}
+		_, resp, err := apiClient.V1BlockGetLatestFinalizedMagicBlock(t, &aa, http.StatusNotModified)
+		require.Equal(t, resp.RawResponse.StatusCode, http.StatusNotModified)
```
stewartie4, line 30: Add the actual output as an extra param to these assertions so if the assertion fails we also get useful output

tests/api_tests/get_latest_finalized_magic_block_test.go
```diff
@@ -0,0 +1,108 @@
...
+		require.Nil(t, err)
+
+		require.Empty(t, string(resp.Body()))
+	})
+
+	t.Run("Not modified, no node lfmb", func(t *testing.T) {
```
stewartie4, line 36: These tests should use the new system tests framework, see https://0chain.slack.com/archives/C02AV6MKT36/p1669414630904799

internal/api/util/client/api.go
```diff
@@ -1131,6 +1134,25 @@ func (c *APIClient) V1BlobberGetFileRefPaths(t *testing.T, blobberFileRefPathReq
 	return blobberFileRefPathResponse, resp, err
 }
 
+func (c *APIClient) V1BlockGetLatestFinalizedMagicBlock(t *testing.T, LFMB *model.LatestFinalizedMagicBlock, requiredStatusCode int) (*block.Block, *resty.Response, error) {
+	var getLFMBResponse *block.Block
+	t.Log("Get latest finalized magic block")
+	urlBuilder := NewURLBuilder().SetPath(GetLatestFinalizedMagicBlock)
+	if LFMB.LFMB.Hash != "" {
+		urlBuilder = urlBuilder.AddParams("node-lfmb-hash", string(LFMB.LFMB.Hash))
```
stewartie4, line 1142: No need for a complex LFMB object if all you require from it is the hash.
I would change this to a string, which makes for easier readability too

stewartie4 (changes requested):

tests/api_tests/get_latest_finalized_magic_block_test.go
stewartie4: @SinanEkm not sure this is resolved?

SinanEkm: @stewartie4 I couldn't exactly understand how it should be done. Is there any example that I can look at?

SinanEkm: @stewartie4 I guess they should be inside each test case right? Now, I changed it like this.

tests/api_tests/get_latest_finalized_magic_block_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/0chain/system_test/internal/api/util/test"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetLatestFinalizedMagicBlock(testSetup *testing.T) {
+	t := test.NewSystemTest(testSetup)
+	t.Parallel()
+
+	t.Run("Lfmb node hash not modified, should return http 304 and empty body", func(t *test.SystemTest) {
+		t.Parallel()
```
Kishan-Dhakan, line 18: When you pass in t *test.SystemTest, it automatically runs test cases in parallel. If you want any test to run in sequence, you need to call t.RunSequentially. Moreover, default case timeout is now 40 seconds. Look these methods for future reference:
image

For now, please remove all t.Parallel() calls, not needed here anymore

Kishan-Dhakan (changes requested):

tests/api_tests/get_latest_finalized_magic_block_test.go
Kishan-Dhakan: Except the top level t.Parallel():

```
t := test.NewSystemTest(testSetup)
t.Parallel()
```
Keep this as it is

Kishan-Dhakan (approved):

Issue #172, opened 2022-01-26
Stake pool unlock and unstaked tokens.


Sriep:
We need a new test to check stake pool lock and unlocking, and unstaking. 

This relates to https://github.com/0chain/0chain/pull/854.

For example something like this
1. Lock tokens in blobber's stake pool.
2. Open allocation. Make sure uses up most of blobber's stake.
3. Get stake pool info, confirm total offer slice increased by a new offer.
4. Lock tokens in blobber's stake pool.
5. Check stake pool
6. Unlock tokens from (4). Should unlock ok.
7. Unlock tokens from (1). Should fail.
8. Get stake pool info. Confirm unlocking tokens amount.
9. Cancel allocation.
8. Unlock tokens from (1), should unlock ok now.
We need to avoid any other test using the blobber and messing up this test. For example, we could add a blobber specifically for this test and add it to the list of blobbers to use for new allocation.

PR #494, opened 2022-12-14
#172 - Stake pool unlock and unstaked tokens

SinanEkm:
It is just a draft PR to see how it will work in a real environment. Because of that, there are too many comment lines and unnecessary variables that will be removed later.


Kishan-Dhakan:
Closing due to no progress @SinanEkm.

PR #491, opened 2022-12-08
Miner block rewards test

Sriep:

PR #501, opened 2022-12-21
Sharder block rewards

Sriep:
Requires https://github.com/0chain/system_test/pull/491


Sriep:
merged with https://github.com/0chain/system_test/pull/491

PR #526, opened 2023-01-28
Remove system test which tests creation of allocation with name

areebusmaniblr:
Removing system tests which tests creation of allocation with name.

In a follow up PR, will remove allocation name from 0chain which will make this tests redundant.
https://github.com/0chain/0chain/pull/2057


Kishan-Dhakan (approved):
LGTM

PR #544, opened 2023-02-12
Task/zs3server test

### LINKED ISSUE #382: API test for zs3server
There is no better place to raise this as an issue, so I am posting it here.

We have [mino-server](https://github.com/0chain/zs3server), which can run a server that is amazon-s3 api compatible and its back-end is mapped to an allocation.
So in summary, s3-server will run a server that can do file operation with s3-like api for an allocation.

Following are the steps for api testing:
1. Create an allocation and lock read-tokens, etc.
2. Run s3-server with credentials set in cli command.
3. We can use minio-client to make varieties of requests and test against the response.
4. We can also make http requests to the api exposed by s3-server. It is more granular and will help other developers to reference for development of program with http requests rather than using with minio-client.

To create above test program we can create some directory in https://github.com/0chain/zs3server, which contains code to test its functionalities.

Please check [this doc](https://github.com/0chain/zs3server/blob/master/docs/gateway/zcn.md) for running s3-server.

ahmedzidan:
- Resolves https://github.com/0chain/system_test/issues/559
- Resolves #382


tests/api_tests/list_objects_test.go
```diff
@@ -0,0 +1,24 @@
...
+	t.Run("ListObjects should return 200 when all the parameters are correct", func(t *test.SystemTest) {
+		queryParams := map[string]string{
+			"accessKey":       AccessKey,
+			"secretAccessKey": SecretAccessKey,
+			"action":          "listObjects",
+			"bucketName":      "root",
+		}
+		resp, err := zs3Client.BucketOperation(t, queryParams, map[string]string{})
+		require.Nil(t, err)
+		require.Equal(t, 200, resp.StatusCode())
```
Kishan-Dhakan, line 22: Add a failing case, ex., listing objects in someone else's bucket should fail

Kishan-Dhakan (changes requested):

PR #549, opened 2023-02-14
miner/sharder block reward changes

Sriep:
Modify miner and sharder block reward test to work with https://github.com/0chain/0chain/pull/2106, which adds provider id to delegate pool rewards.

Requires https://github.com/0chain/0chain/pull/2106


peterlimg (approved):
lgtm

PR #551, opened 2023-02-14
gosdk to staging

devyetii:
I was facing a problem with an api test [here](https://github.com/0chain/system_test/blob/68cc37e00909f6958921693bc2c084f07314a2bc/tests/api_tests/get_blobber_rewards_test.go#L105) then I realized gosdk is used directly in this system test, however, it was pointing to a pretty old version of gosdk. Here I upgrade it to gosdk's staging. I believe it's not the problem with the test failing with me, but I think it should be kept up-to-date with staging.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/0chain/errors v1.0.3
-	github.com/0chain/gosdk v1.8.10-0.20221013065231-bc842616c33e
+	github.com/0chain/gosdk v1.8.13-0.20230213081016-7ba75bc773f2
 	github.com/go-resty/resty/v2 v2.7.0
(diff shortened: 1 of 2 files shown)


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/gosdk-to-staging  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | 1927-revisit-all-evens-that-make-upsert  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #556, opened 2023-02-15
increase timeout

Kishan-Dhakan:


stewartie4 (approved):
Approved for now but not ideal, we should be increasing the timeout on the slow test, not every test

PR #622, opened 2023-04-06
system-test-534

shalinikum:


Kishan-Dhakan:
Where are the new tests in this PR? This only seems to be removing a lot of tests

tests/api_tests/0box_test.go
```diff
@@ -16,6 +16,673 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// Keeping at top otherwise it would not work
+func Test0BoxAllocation(testSetup *testing.T) {
+	// todo: These tests are sequential and start with teardown as they all share a common phone number
```
stewartie4, line 21: Why move these tests to a single file, is it not tidier as separate files?

stewartie4 (changes requested):

tests/api_tests/0box_test.go
shalinikum: Hi Ryan,
we don't know in which order the test file would run, and there are lots of other testcase which are using allocation endpoint. Keeping that in mind we should move allocation at top.

Kishan-Dhakan (approved):
LGTM

tests/api_tests/0box_test.go
Kishan-Dhakan: @shalinikum it was added to a new file for the exact reason you mentioned, we know `0box_allocation_test.go` will run before `0box_tests.go`, since it is lexicographically smaller name. I'v reverted the separation to as it was.

shalinikum: okay.Kishan Thank you so much. I didn't new that.

stewartie4 (approved):
LGTM - good increased coverage

Kishan-Dhakan (approved):
LGTM

Issue #630, opened 2023-04-12
System Test for kill miner and kill sharder


Sriep:
Requires https://github.com/0chain/0chain/pull/2285 https://github.com/0chain/gosdk/pull/891 and  https://github.com/0chain/zwalletcli/pull/154

PR #631, opened 2023-04-12
System test for kill_miner and kill_sharder

Sriep:
closes https://github.com/0chain/system_test/issues/630 

Requires

0chain: https://github.com/0chain/0chain/pull/2285
goSDK: https://github.com/0chain/gosdk/pull/891
zwallet: https://github.com/0chain/zwalletcli/pull/154


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | kill_miner  |
| gosdk  | minersc_kill  |
| zboxcli  | staging  |
| zwalletcli  | kill_miners  |
| 0chain  | pr-2285-0e9c3888  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | kill_miner  |
| gosdk  | minersc_kill  |
| zboxcli  | staging  |
| zwalletcli  | kill_miners  |
| 0chain  | kill_minersc  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

tests/cli_tests/zwalletcli_kill_sharder_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+		output, err := killSharder(t, escapedTestName(t), configPath, createParams(map[string]interface{}{
+			"id": sharderIds[0],
+		}), true)
+		require.Error(t, err, "kill sharder by non-smartcontract owner should fail")
+		require.Len(t, output, 1)
+		require.True(t, strings.Contains(output[0], "unauthorized access - only the owner can access"), "")
+	})
+
+	t.RunSequentiallyWithTimeout("Killed sharder does not receive rewards", 1000*time.Second, func(t *test.SystemTest) {
```
Kishan-Dhakan, line 46: Why the substantially longer timeout here, 1000 is 5x of shutdown miner

Kishan-Dhakan (changes requested):

tests/cli_tests/zwalletcli_kill_sharder_test.go
Sriep: The longer timeout here is that there are often no sharders available to kill, so we must wait for one to be added to the MPT. For some reason, there is usually a miner we can kill registered, so the wait is not so important.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | kill_miner  |
| gosdk  | minersc_kill  |
| zboxcli  | kill_miners  |
| zwalletcli  | staging  |
| 0chain  | kill_minersc  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

stewartie4 (commented):
is it possible that these tests will affect other tests? eg. by killing a miner/validator?

Kishan-Dhakan:
> is it possible that these tests will affect other tests? eg. by killing a miner/validator?
Yep, the nodes being killed can't be used again. I'll make the necessary CI changes today to close this.

peterlimg:
@Kishan-Dhakan what's the plan on CI to fix this issue, maybe we should run the kill and shutdown tests in the end of the tests so that it would not affect other cases.

Sriep:
I added a check that there are at least a minimum number of miners/sharders in the 0chian. The tests are skipped if there are not enough miners/sharders. 

The theory is that if we have zero or one miner/sharder in the 0chain then the 0chain is not suitable for this test; as the system tests should be able to run against any valid 0chian the solution would seem to be skipping the test in this situation. If there is still a problem with other tests we can increase the number of miners/sharders until there is not.

Adding a framework to allow partial ordering of tests would be a nice addition.

PR #636, opened 2023-04-15
Edited path to the correct location for running tests locally

mayank-2105:
The location to copy zbox cli and zwallet clis was set to tests/cli/ while there is no such directory in system_test repo. Edited it to tests/cli_tests directory


Kishan-Dhakan (approved):
Good catch, LGTM

stewartie4 (approved):
Thanks good catch

PR #639, opened 2023-04-17
Added tokenomics code to new workflow

Jayashsatolia403:


.github/workflows/ci.yml
```diff
@@ -172,8 +172,8 @@ jobs:
           deploy_report_page: true
           archive_results: true
           run_flaky_tests: true
-          run_api_system_tests: true
-          run_cli_system_tests: true
+          run_api_system_tests: false
+          run_cli_system_tests: false
```
Kishan-Dhakan, line 176: This ci.yml is the workflow file for manual runs and automated runs on every commit, we don't want to set them to false here. What you want to do is, create a separate workflow file, that runs manually, and only runs tokenomics suite. You can read on github workflows or check out or workflow files to see how they work.

Kishan-Dhakan (changes requested):
Partial review: we don't want to turn off other tests in the main runs, we want to keep tokenomics suite disabled in the main run, and run only tokenomics suite when we need (via a manual trigger or cron job)

.github/workflows/ci.yml
Jayashsatolia403: Yeah, I just turned them off to get check if my changes are working faster.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | tokenomics_suite  |
| gosdk  | staging  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | tokenomics_suite_cicd_test  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

PR #660, opened 2023-05-04
Miner and sharder fee rewards

Sriep:
Reactivate tests after 0chain fix in https://github.com/0chain/0chain/pull/2338

This does not work at the moment as, due to 0chain issues, the event database transaction table is not always populated correctly. This test makes use of the transaction table to calculate transaction fees. Hence this is put on hold, waiting for 0chain fixes.


stewartie4 (commented):
@Sriep are we good to merge this?

PR #662, opened 2023-05-06
Added changes to make it compatible with the new time unit

Jayashsatolia403:


tests/cli_tests/zboxcli_cancel_allocation_test.go
```diff
@@ -79,15 +79,15 @@ func TestCancelAllocation(testSetup *testing.T) {
 		require.Equal(t, "Error creating allocation:alloc_cancel_failed: value not present", output[0])
 	})
 
-	t.Run("Cancel Expired Allocation Should Fail", func(t *test.SystemTest) {
+	t.RunWithTimeout("Cancel Expired Allocation Should Fail", 12*time.Minute, func(t *test.SystemTest) {
```
stewartie4, line 82: so we're just waiting 12 minutes for the allocation to expire? this isn't workable, our dev CI needs to be fast, why can we not configure for a faster expity?

tests/cli_tests/zboxcli_finalize_allocation_test.go
```diff
@@ -19,7 +19,7 @@ func TestFinalizeAllocation(testSetup *testing.T) {
 
 	t.Parallel()
 
-	t.RunWithTimeout("Finalize Expired Allocation Should Work after challenge completion time + expiry", 10*time.Minute, func(t *test.SystemTest) {
+	t.RunWithTimeout("Finalize Expired Allocation Should Work after challenge completion time + expiry", 17*time.Minute, func(t *test.SystemTest) {
```
stewartie4, line 22: exactly what i mean

stewartie4 (changes requested):
we need to re-think how to do this on dev CI

stewartie4 (commented):
My ask is that we reduce the time to the lowest number possible that works (should that be 1 minute or 5 minutes), and update the sleeps/timeouts to match that, right now many tests are sleeping for too long

PR #668, opened 2023-05-11
Fix unstake tokens test

devyetii:


stewartie4 (approved):
Happy to approve pending green tests, but I'm not sure this is the fix for the issue - the chain seems to die whereas this is a simple assertion change

stewartie4 (approved):
Happy to approve pending green tests, but I'm not sure this is the fix for the issue - the chain seems to die whereas this is a simple assertion change

PR #670, opened 2023-05-12
Add Sleep after upload 

Hitenjain14:
We need to wait for some time after uploading the file, so blobber can redeem the writemarker.

### DIFF
--- a/tests/cli_tests/zboxcli_file_move_test.go
+++ b/tests/cli_tests/zboxcli_file_move_test.go
@@ -271,7 +271,7 @@ func TestFileMove(testSetup *testing.T) { // nolint:gocyclo // team preference i
 		localpath := uploadRandomlyGeneratedFile(t, allocationID, "/", fileSize)
 
 		// Wait for write marker to be redeemed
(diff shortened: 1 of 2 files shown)


stewartie4 (commented):
The tests above don't verify that the write marker has been redeemed so are these waits necessary? And why 16 seconds?

Hitenjain14:
> The tests above don't verify that the write marker has been redeemed so are these waits necessary? And why 16 seconds?

These tests check if the amount has been deducted from the write pool or not on update, for this we check the difference between initial allocation and final allocation write pool. We have no control of when these write markers are redeemed, so we wait for 16 seconds for them to be redeemed so that the initial balance is of after the tokens has been deducted.

Hitenjain14:
If u don't wait you can get a value of the write pool where no blobbers have redeemed the writemarker or 1 of them and soo on.

stewartie4 (approved):
Ok thanks for the tests update, happy to merge once they are passing

dabasov (approved):
lgtm

PR #689, opened 2023-05-22
Fix lock negative tokens

Jayashsatolia403:


Kishan-Dhakan (changes requested):
I see it's avoiding mustFailCost from going to negative. But, negative cost should fail as well. Since we don't have another case for negative cost must fail, please add that case.

tests/cli_tests/zboxcli_create_allocation_test.go
```diff
@@ -73,10 +73,29 @@ func TestCreateAllocation(testSetup *testing.T) {
...
+			"cost":        "",
+			"read_price":  "0-1",
+			"write_price": "0-1",
+			"size":        10000,
+			"expire":      "30m",
+		}
+		mustFailCost := -1
+		options = map[string]interface{}{"lock": mustFailCost}
+		output, err := createNewAllocationWithoutRetry(t, configPath, createParams(options))
+		require.NotNil(t, err, strings.Join(output, "\n"))
```
Kishan-Dhakan, line 98: Could've used a //TODO comment to fix this assertion once 2431 is fixed

Kishan-Dhakan (approved):

PR #733, opened 2023-06-23
Add/transaction tests

lpoli:
This PR add system-tests to test some cases for putting transaction into miner's transaction pool.
It resolves https://github.com/0chain/system_test/issues/744 issue


tests/api_tests/main_test.go
```diff
@@ -40,7 +43,17 @@ func TestMain(m *testing.M) {
...
 	zs3Client = client.NewZS3Client(parsedConfig.ZS3ServerUrl)
 	zboxClient = client.NewZboxClient(parsedConfig.ZboxUrl, parsedConfig.ZboxPhoneNumber)
+	configMap := map[string]interface{}{
+		"block_worker":              parsedConfig.BlockWorker,
+		"signature_scheme":          "bls0chain",
+		"min_submit":                50,
+		"min_confirmation":          50,
+		"confirmation_chain_length": 3,
+		"max_txn_query":             5,
+		"query_sleep_time":          5,
```
Kishan-Dhakan, line 53: This should be in a config file and taken from there. Can abstract this into a method that returns string(b) for the arg of zcncore.Init() method.

Kishan-Dhakan (changes requested):
Partial review

stewartie4 (approved):
Looks good, can we get multiple runs of the new tests before merging?

PR #755, opened 2023-07-05
fix assertion after https://github.com/0chain/s3-migration/pull/79

Kishan-Dhakan:


Kishan-Dhakan:
Merged with https://github.com/0chain/s3-migration/pull/79

PR #784, opened 2023-07-19
test free download and authticket download

Kishan-Dhakan:


tests/cli_tests/0_free_read_test.go
```diff
@@ -0,0 +1,168 @@
...
+	})
+
+	t.RunSequentially("Free reads should work", func(t *test.SystemTest) {
+		output, err := createWallet(t, configPath)
+		require.Nil(t, err, "Failed to create wallet", strings.Join(output, "\n"))
+
+		_ = setupWallet(t, configPath)
+
+		// Create an allocation
+		options := map[string]interface{}{"size": 1 * MB, "lock": "0.5"}
```
dabasov, line 70: can we add readPrice here?

Kishan-Dhakan: Is it needed? All blobbers have 0 read price now so default range is fine

dabasov: if price changes to blobbers eventually, we can have an error early, but not have a false positive test with not free read

tests/cli_tests/0_free_read_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// revert read prices irrespective of test results
+	t.Cleanup(func() {
+		for _, blobber := range blobberList {
+			output, err := updateBlobberInfo(t, configPath, createParams(map[string]interface{}{"blobber_id": blobber.ID, "read_price": intToZCN(blobber.Terms.Read_price)}))
+			require.Nil(t, err, strings.Join(output, "\n"))
+		}
+	})
+
+	t.RunSequentially("Free reads should work", func(t *test.SystemTest) {
```
stewartie4, line 63: Out of interest why is this sequential but the other test parallel?

tests/cli_tests/0_free_read_test.go
```diff
@@ -0,0 +1,168 @@
...
+		}), true)
+		require.Nil(t, err, strings.Join(output, "\n"))
+		require.Len(t, output, 2)
+
+		require.Contains(t, output[1], StatusCompletedCB)
+		require.Contains(t, output[1], filepath.Base(filename))
+
+		downloadedFileChecksum := generateChecksum(t, "tmp/"+filepath.Base(filename))
+
+		require.Equal(t, originalFileChecksum, downloadedFileChecksum)
```
stewartie4, line 166: nice

stewartie4 (approved):
Great coverage

stewartie4:
tests not running because it looks like the "free-reads" branch doesn't exist in blobber repo

dabasov (approved):
lgtm

tests/cli_tests/0_free_read_test.go
Kishan-Dhakan: Nice catch, both should be parallel

Issue #336, opened 2022-07-23
Enhance Challenge Tests


Kishan-Dhakan:
- Enhance challenge tests to assert challanges are passing under mcct.
- Sync with blobber team to see what check are needed on bobber's data for challenge to pass.
- Current tests use old logic of CRUD operation for challenge generation. This needs to be removed completely and no. of challenge generated should be equal to round number.

PR #790, opened 2023-07-22
Fix challenge system tests

Jayashsatolia403:


stewartie4 (changes requested):
@Jayashsatolia403 your recent commit seems to have wiped this branch 
And as it was a rebase, the history is also gone 

Can this be restored and in future can you just use regular commits?
Commit history is really useful when reviewing a PR

Kishan-Dhakan (approved):
Can be improved a lot more, see https://github.com/0chain/system_test/issues/336

PR #804, opened 2023-07-29
Unskip s3 tests

Kishan-Dhakan:


tests/cli_tests/0_s3mgrt_migrate_test.go
```diff
@@ -71,7 +71,7 @@ func Test0S3Migration(testSetup *testing.T) {
 			"allocation": allocationID,
 		}))
 
-		require.NotNil(t, err, "Expected a migration failure but got no error", strings.Join(output, "\n"))
+		require.Nil(t, err, "Expected a Migration completed successfully", strings.Join(output, "\n"))
```
Kishan-Dhakan, line 74: This is not correct, why would you not expect a failure if allocation is smaller than what is in the bucket?

PR #823, opened 2023-08-13
Unksip zs3 system tests

Hitenjain14:


Kishan-Dhakan (approved):
Lgtm, gosdk is also sprint-july-4

PR #831, opened 2023-08-19
Fix/reduce allocation

Jayashsatolia403:


tests/cli_tests/zboxcli_update_allocation_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
Kishan-Dhakan, line 16: Where is assert being used?

Kishan-Dhakan (changes requested):

Kishan-Dhakan (approved):

PR #844, opened 2023-08-27
Fix long file names in case of duplicates

sunilmhta:
Fix long file names in case of duplicates

Related PR:
s3-migration: https://github.com/0chain/s3-migration/pull/96


tests/cli_tests/0_s3mgrt_migrate_alternate2_test.go
```diff
@@ -400,7 +400,9 @@ func Test0S3MigrationAlternatePart2(testSetup *testing.T) {
 // This func needs to be in sync with the original func.
 func getUniqueShortObjKey(objectKey string) string {
 	// Max length to which objectKey would be trimmed to.
-	const maxLength = 100
+	// Keeping this less than 100 chars to prevent longer name in case of uploading duplicate
```
stewartie4, line 403: is there a test to verify what happens on filenames greater than 100 chars?

PR #865, opened 2023-09-24
add workflow

Kishan-Dhakan:


Kishan-Dhakan:
Merging, needed to run the workflow

PR #887, opened 2023-10-11
remove killed/shutdown blobber

Kishan-Dhakan:


Kishan-Dhakan:
Passing run: https://github.com/0chain/system_test/actions/runs/6483682791/job/17605829194

PR #919, opened 2023-10-29
Feature/multi alloc stats

Jayashsatolia403:


tests/api_tests/0box_aggregate_endpoints_test.go
```diff
@@ -51,6 +51,313 @@ func Test0boxGraphAndTotalEndpoints(testSetup *testing.T) {
 	marker := config.CreateFreeStorageMarker(t, sdkWallet.ToSdkWallet(sdkWalletMnemonics), ownerWallet.ToSdkWallet(ownerWalletMnemonics))
 	t.Logf("Free allocation marker: %v", marker)
 
+	t.RunSequentiallyWithTimeout("test multi allocation overall graph data", 10*time.Minute, func(t *test.SystemTest) {
+		// Get initial total challenge pools
```
stewartie4, line 55: My only comment is this seems to be a very complex test - are we sure it is not flaky and do we have enough logging to debug it if it fails?

Jayashsatolia403: Yes, I have run it multiple times to check if it works. @dabasov wanted to add a test that covers every param in snapshot when we have multiple allocations and file operations on network. 

This test will be running for roughly 4 to 5 minutes.

PR #930, opened 2023-11-05
Updated chimney blobber rewards to work with 0 service charge

Jayashsatolia403:


tests/api_tests/chimney_blobber_rewards_test.go
```diff
@@ -179,13 +179,27 @@ func Test1ChimneyBlobberRewards(testSetup *testing.T) {
...
+			t.Log("Blobber ID: ", blobber.ID)
+			t.Log("Blobber Challenge Count: ", blobberChallengeCount)
 
 			// Updating total values
 			actualCancellationCharge += actualCancellationChargeForBlobber
 
 			expectedCancellationChargeForBlobber := expectedCancellationCharge * writePriceWeight(blobber.Terms.WritePrice, totalWritePrice)
+			expectedCancellationChargeForBlobber *= float64(blobberChallengeCount["passed"]+blobberChallengeCount["open"]) / float64(blobberChallengeCount["total"])
+
+			t.Log("Expected Challenge Reward: ", expectedCancellationChargeForBlobber)
```
stewartie4, line 191: Remove all of the t.log in this file as the assertion will print out the expected and actual if it fails

stewartie4 (changes requested):

stewartie4:
small tweak needed, otherwise looks good

tests/api_tests/chimney_blobber_rewards_test.go
Jayashsatolia403: It does not in case of inepsilon.

stewartie4 (approved):

PR #933, opened 2023-11-07
Feature/multi format file upload

Jayashsatolia403:


tests/cli_tests/zboxcli_file_upload_test.go
```diff
@@ -299,6 +301,58 @@ func TestUpload(testSetup *testing.T) {
 		require.Equal(t, expected, output[1])
 	})
 
+	t.RunWithTimeout("Upload tests with Thumbnail with different format", 40*time.Minute, func(t *test.SystemTest) {
+		for _, blobberId := range blobbersList {
+			_, err := executeFaucetWithTokens(t, configPath, 11)
+			require.Nil(t, err, "Error executing faucet")
+
+			t.Log("Staking tokens for blobber: ", blobberId)
```
stewartie4, line 309: can be removed

tests/cli_tests/zboxcli_file_upload_test.go
```diff
@@ -299,6 +301,58 @@ func TestUpload(testSetup *testing.T) {
...
+			_, err = stakeTokens(t, configPath, utils.CreateParams(map[string]interface{}{
+				"blobber_id": blobberId,
+				"tokens":     10,
+			}), true)
+			require.Nil(t, err, "Error staking tokens")
+		}
+
+		allocSize := int64(10 * GB)
+
+		var fileExtensions = []string{".txt", ".docx", ".pdf", ".jpg", ".png", ".mp3", ".mp4", ".xlsx", ".html", ".json", ".csv", ".xml", ".zip", ".rar", ".gz", ".tar", ".avi", ".mov", ".wav", ".ogg", ".bmp", ".gif", ".svg", ".tiff", ".ico", ".py", ".c", ".java", ".php", ".js", ".css", ".scss", ".yaml", ".sql", ".md", ".go", ".rb", ".cpp", ".h", ".sh", ".bat", ".dll", ".class", ".jar", ".exe", ".psd", ".pptx", ".xls", ".ppt", ".key", ".numbers"}
```
stewartie4, line 321: do we need all of these? it's never going to be an extensive list, so why not make it shorter so the test takes less time to run>

tests/cli_tests/zboxcli_file_upload_test.go
```diff
@@ -299,6 +301,58 @@ func TestUpload(testSetup *testing.T) {
...
+				"thumbnailpath": thumbnail,
+			}, true)
+			require.Nil(t, err, strings.Join(output, "\n"))
+			require.Len(t, output, 2)
+
+			expected := fmt.Sprintf(
+				"Status completed callback. Type = application/octet-stream. Name = %s",
+				filepath.Base(filename),
+			)
+			require.Equal(t, expected, output[1], "Failed to upload file with extension: "+ext)
```
stewartie4, line 352: The actual output should be returned in the message if the assertion fails

stewartie4 (changes requested):

tests/cli_tests/zboxcli_file_upload_test.go
Jayashsatolia403: Why do we want to remove staking?

tests/cli_tests/zboxcli_file_upload_test.go
Jayashsatolia403: Yes, we do.

tests/cli_tests/zboxcli_file_upload_test.go
Jayashsatolia403: Fixing. Thanks!!

tests/cli_tests/zboxcli_file_upload_test.go
stewartie4: We don't want to remove staking, we want to remove the superfluous log line

PR #941, opened 2023-11-12
Fix tokenomics nightly runs

Jayashsatolia403:


.github/workflows/tokenomics_ci.yml
```diff
@@ -46,9 +46,9 @@ jobs:
           echo "TOKENOMICS_BLOCK_REWARD_TESTS=$(echo $(([ -z 'TestBlockRewardsForBlobbers' ] && echo '') || echo 'TestBlockRewardsForBlobbers'))" >> $GITHUB_ENV
           echo "TOKENOMICS_MIN_STAKE_TESTS=$(echo $(([ -z 'TestMinStakeForProviders' ] && echo '') || echo 'TestMinStakeForProviders'))" >> $GITHUB_ENV
           echo "CLIENT_THROTTLING=$(echo $(([ -z 'TestClientThrottling' ] && echo '') || echo 'TestClientThrottling'))" >> $GITHUB_ENV
-          echo "CURRENT_BRANCH=$(echo $(([ -z '${{ github.head_ref }}' ] && echo  ${GITHUB_REF#refs/*/}) || echo '${{ github.head_ref }}'))" >> $GITHUB_ENV
-          echo "REPO_SNAPSHOTS_BRANCH=$(echo $(([ -z '${{github.event.inputs.repo_snapshots_branch}}' ] && echo '<CURRENT BRANCH WITH FALLBACK TO CURRENT_SPRINT>') || echo '${{github.event.inputs.repo_snapshots_branch}}'))" >> $GITHUB_ENV
-
+          echo "CURRENT_BRANCH=sprint-1.11" >> $GITHUB_ENV
+          echo "REPO_SNAPSHOTS_BRANCH=current-sprint" >> $GITHUB_ENV
```
stewartie4, line 50: This is the wrong approach.
The existing logic took the repo snapshots branch from user input and used that. If no branch was selected, it fell back to the current sprint.
We shouldn't be hardcoding runs to a specific sprint or branch as they will diverge quickly
what are you trying to do?

stewartie4 (changes requested):

.github/workflows/tokenomics_ci.yml
Jayashsatolia403: We want nightly runs on sprint branches. How should I achieve that?

PR #949, opened 2023-11-18
Feature/Upload 50 large and 50 small files for different formats with multioperation till allocation size (of 2gb)

Jayashsatolia403:


tests/api_tests/multiop_test.go
```diff
@@ -33,7 +33,7 @@ func TestMultiOperation(testSetup *testing.T) {
 	t.SetSmokeTests("Multi upload operations should work")
 
 	t.RunSequentially("Multi upload operations should work", func(t *test.SystemTest) {
-		createAllocationAndPerformMultiOperation(t, 1*KB, 10, 10, false, []int64{}, "")
+		createAllocationAndPerformMultiOperation(t, 0, 10, 10, false, []int64{}, "")
```
stewartie4, line 36: why are we zeroing this out?

tests/api_tests/challenge_timings_test.go
```diff
@@ -95,7 +95,7 @@ func TestProtocolChallengeTimings(testSetup *testing.T) {
 		require.NoError(t, err)
 
 		fileSize := int64(1 * MB)
-		uploadOp := sdkClient.AddUploadOperation(t, allocationID, fileSize)
+		uploadOp := sdkClient.AddUploadOperation(t, "", fileSize)
```
stewartie4, line 98: why are we blanking this out?

tests/api_tests/multiop_test.go
```diff
@@ -355,3 +276,67 @@ func randName() string {
 	}
 	return string(b)
 }
+
+func createAllocationAndPerformMultiOperation(t *test.SystemTest, allocSize int64, filesCount, expectedFilesCount int, fileWithFormats bool, fileSizes []int64, secondaryOperation string) {
```
stewartie4, line 280: can this be separated into 2 functions?

tests/api_tests/multiop_test.go
```diff
@@ -355,3 +276,67 @@ func randName() string {
...
+		}
+
+		start := time.Now()
+		sdkClient.MultiOperation(t, allocationID, newOps)
+		end := time.Since(start)
+		t.Logf("Multi "+secondaryOperation+"operations took %v", end)
+	}
+
+	listResult := sdkClient.GetFileList(t, allocationID, "/")
+	require.Equal(t, expectedFilesCount, len(listResult.Children), "files count mismatch expected %v actual %v", 10, len(listResult.Children))
```
stewartie4, line 341: How stable is this test? has it been ran successfully multiple times?

Jayashsatolia403: Here is full green run. I tried 2 times and it works : https://github.com/0chain/system_test/actions/runs/6916741139/job/18817133384. 

Let's merge and see if it fails on nightly runs.

tests/api_tests/multiop_test.go
Jayashsatolia403: It just sets default value when we pass 0 in input.

tests/api_tests/multiop_test.go
Jayashsatolia403: This can be separated into multiple functions but I think this is already a single use case so let's keep it in single function. 

I just wanted to avoid rewriting all the code again and again so moved complete changes into a different function.

tests/api_tests/challenge_timings_test.go
Jayashsatolia403: This is different param now. I removed allocationID filed and added new function parameter format. 

Empty string means we don't need any specific file format.

stewartie4 (approved):

PR #964, opened 2023-12-03
Fix sprint and resolve conflict to merge master

Jayashsatolia403:


stewartie4 (approved):
Happy to merge as everything has been approved previously, but we'll definitely need a green build before merging

PR #1096, opened 2024-07-25
running rollback allocation tests parallel.

yash10019coder:
Instead of using `RunTestSequential` using `RunTest` to run them parallel.


Jayashsatolia403 (approved):
lgtm

PR #1113, opened 2024-10-06
fixes for gdrive and dropbox migration tests

akash-singh8:
for typo fix, it should be access-key, run `./s3migration migrate --help` to verify
```
Usage:
  s3migration migrate [flags]

Flags:
      --access-key string      access-key of CloudStorage [Aws, Drive, Dropbox]
```


pewssh:
I have pulled changes in another branch https://github.com/0chain/system_test/tree/dropbox_drive_migration

PR #1118, opened 2024-10-15
system test for new transaction endpoint

smaulik13:


tests/api_tests/0box_transactions._test.go
```diff
@@ -0,0 +1,26 @@
...
+	"time"
+
+	"github.com/0chain/system_test/internal/api/util/test"
+	"github.com/stretchr/testify/require"
+)
+
+func Test0BoxTransactions(testSetup *testing.T) {
+	t := test.NewSystemTest(testSetup)
+
+	t.RunSequentiallyWithTimeout("get paginated transactions list while creating pit id", 5*time.Minute, func(t *test.SystemTest) {
```
Jayashsatolia403, line 14: Why 5 minutes timeout?

tests/api_tests/0box_transactions._test.go
```diff
@@ -0,0 +1,26 @@
...
+		txnData, resp, err := zboxClient.GetTransactionsList(t, "")
+		require.NoError(t, err)
+		require.Equal(t, 200, resp.StatusCode())
+
+		if txnData != nil {
```
Jayashsatolia403, line 19: We will always have transactions so txnData should never be nil. Return error if it is nil.

tests/api_tests/0box_transactions._test.go
```diff
@@ -0,0 +1,26 @@
...
+			pitId := txnData.PitId
+			_, resp, err := zboxClient.GetTransactionsList(t, pitId)
```
Jayashsatolia403, line 21: Verify the data returned as well. You can request the same transactions from 0chain by hash and compare the content.

Jayashsatolia403 (changes requested):

tests/api_tests/0box_transactions._test.go
smaulik13: As once pit id is created it is valid for 5 mins

Jayashsatolia403: It's timeout for the test run. You need to know how much should it take to run.

Jayashsatolia403 (commented):
lgtm

